Issue #14, opened 2020-09-19
Qubit allocation without new block scope


samarsha:
**Suggestion**

We propose creating an analogue of `let` for qubits: a statement that allocates qubits and binds them to a name for the remainder of the current scope, after which the name is no longer accessible and the qubits are released. The statement does not require a new block scope, unlike the current `using` and `borrowing` statements.

*Considerations*

Allocating qubits in Q# with `using` and `borrowing` statements requires creating a new block scope. The block makes the lifetime of the qubits explicit, and discourages holding on to qubits for longer than needed, but at the cost of increased block nesting and indentation. Additionally, it is very common for a `using` statement block to be the last statement in the enclosing scope, in which case the explicit scope did not actually help to shorten the lifetime of the allocated qubits.

The explicit block also makes it cumbersome to create more than one qubit variable at a time. If one `using` statement per variable is used, a new indentation level is required for each variable. More commonly, tuple destructuring is used to create more than one variable in a single assignment, but this becomes difficult to read as the number of variables increases or as the complexity of the qubit initializer expression increases.

A `let`-like qubit allocation statement is needed to simplify these use cases. Because `using` and `borrowing` statements are the only way to allocate qubits, there is no workaround currently in Q#. An alternative proposal that would only address the many variable case could be to allow multiple `using` statements in a row followed by only one scope:

```qsharp
using (q1 = Qubit())
using (q2 = Qubit()) {
    // ... use both q1 and q2 ...
}
```

But this still requires at least one new block scope in all cases, unlike the original suggestion.

*Context*

We can consider two possible naming conventions for the new statements:

1. Reuse the same keywords as the block statements: `using` and `borrowing`.
2. By analogy to `let`, create the new keywords `use` and `borrow`.

Reusing the keywords avoids a breaking change, but creates a slight inconsistency:

```qsharp
using q1 = Qubit();
borrowing q2 = Qubit();
letting q3 = q1; // wrong
let q3 = q1;     // right
```

If we use new keywords, we need to decide if we keep the original keywords for the block statements (supporting two parallel sets of keywords indefinitely), or if we rename those too:

```qsharp
use (q = Qubit()) {
    // ...
}
borrow (q = Qubit()) {
    // ...
}
```

*Examples*

Example 1: Current syntax.
```qsharp
// One variable.
using (q = Qubit()) {
    // ...
}

// Two variables with separate statements.
using (q1 = Qubit()) {
    using (q2 = Qubit()) {
        // ...
    }
[...]

PR #33, opened 2020-09-28
Proposal for simplified qubit allocation syntax

samarsha:
This is the proposal for issue #14: Qubit allocation without new block scope.


Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+description: Provides more convenient syntax for allocating qubits.
+author: Sarah Marshall
+date: September 28, 2020
+---
+
+# Proposal
+
+1. Add new `use` and `borrow` statements to allocate qubits that are valid until the end of the current block.
+2. Add new scope statement that does not require a preceding keyword, and is executed exactly once; equivalent to `if (true) { }`.
+3. Deprecate and remove existing `using` and `borrowing` block statements.
```
cgranade, line 12: Not strictly related to this proposal, but would it be reasonable to offer a code action or other IDE support to help transition to the new proposed syntax?

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+
+# Justification
+
+Qubit allocation is currently awkward in Q# for two reasons:
+
+1. It requires creating a new block to define the lifetime of the qubit, even when the intended lifetime is the same as the current block.
+2. Allocating multiple qubits at the same time requires either tuple destructuring or nested blocks, both of which can be hard to read.
```
cgranade, line 19: Absolutely agreed, especially because it often feels like qubit allocation in `using` blocks is "transposed" for more complicated allocations:
```
using ((a, b, c) = (Qubit(), Qubit(), Qubit())) {
    // ...
}
```
Never mind that requiring `)))` for a relatively common case feels a bit heavy-weight.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+Qubits are a managed resource in Q#: when qubits are allocated, they are valid only within the new block declared at allocation, after which they are released and can no longer be used.
+This proposal makes qubit allocation in Q# always scoped to the current block.
+Instead of being released at the end of a newly-declared block, qubits are released at the end of the current block.
+It also extends block statement syntax to allow precise control of qubit lifetime when needed in more complicated scenarios.
+Finally, this proposal aims to be syntactically minimal and consistent with other binding statements like `let`, in contrast to alternatives that have more complicated syntax.
+
+## Current Status
+
+Currently, the only mechanism to allocate qubits in Q# is with `using` and `borrowing` block statements.
+These statements will be deprecated and removed, replaced with new `use` and `borrow` statements that do not require a block, as well as a new scope statement that can be used when a new block is desired.
```
cgranade, line 35: Since `use` and `borrow` can currently be variable names (though hopefully they aren't terribly common!), is this a breaking change? Would it make sense to have a opt-in to the new behavior (i.e.: a special Python-like `open` statement, a C#- or Rust-like `#` statement, or similar)?

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+    H(q);
+    return M(q); // q is released here.
+}
+```
+
+Example 2b: Many qubits allocated in a single `use` statement with tuple destructuring.
+
+```qsharp
+operation QubitTuple(n : Int) : Result {
+    use (a, b, c) = (Qubit(), Qubit[2 * n + 1], Qubit[n]);
```
cgranade, line 203: Is destructuring required by this proposal, or do existing tuple deconstruction rules apply? In particular, is the following valid?
```Q#
use pair = (Qubit(), Qubit());
PrepareEntangledPair(pair);
```

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+Instead of adding non-block statements, Q# could extend the existing block statement syntax to allow multiple `using` and `borrowing` statements in a row that have only one block attached:
+
+```qsharp
+using (a = Qubit())
+borrowing (b = Qubit())
+using (c = Qubit()) {
+    // ...
+}
+```
+
```
cgranade, line 317: If I recall, there was a third alternative discussed at one point, where the `use` keyword would appear on the right-hand side:
```Q#
let register = use Qubit[2 * n + 1];
```
It may be helpful to compare to that alternative as well, especially as I think the proposed syntax presents some real advantages over putting `use` on the right.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+Example 5b: A qubit allocated in new block and released before the end of the outer block.
+
+```qsharp
+operation DifferentLifetime() : Result {
+    mutable r = Zero;
+    {
+        use q = Qubit();
```
cgranade, line 238: The one advantage I see to the previous approach over this example is that it's obvious when looking at the start of the block that the block is intended to delineate a qubit lifetime. Though it would be bad style to do so, I could imagine someone writing code like the following, in which case it would be harder to identify the qubit lifetime:

```Q#
mutable r = Zero;
{
     let foo = f(r);
     set r = bar(foo);
     use q = Qubit();
}
```
In that case, it's not strictly _wrong_ to put classical precomputation inside the nested `{}`, but it does make it harder to read what the purpose of the new scope is.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+It does not change the behavior of qubit management; it only provides new syntax for existing functionality.
+
+## Impact on Existing Mechanisms
+
+The scope statement added in this proposal can also be used to limit the scope of non-qubit variable bindings declared with `let` and `mutable`.
+This may be useful for limiting the lifetime of large data structures like arrays or for otherwise communicating the intent that a variable should only be used in a specific section within a callable.
+
+## Anticipated Interactions with Future Modifications
+
+Because the proposed modification is similar to existing functionality, it is not expected to have any impact on future language development.
```
cgranade, line 282: Would it make sense to expand on how the current and proposed behavior would each look when combining with other potential new features like allocating qubits directly into UDTs like `LittleEndian` and `FixedPoint`? For example, would we want to support syntax like `use qs = LittleEndian(Qubit[10]);` or `let qs = LittleEndian(use Qubit[10]);`?

Similarly, how would this proposal look with apply-and-return expressions? I recall you mentioning that something like `use q = Qubit() <~ H;` could be a reasonable way to prepare a qubit in the |+⟩ state, for example.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+}
+
+borrowing q = Qubit();
+// ...
+
+borrowing (q = Qubit()) {
+    // ...
+}
+```
+
```
cgranade, line 305: Perhaps an odd idea on how to adapt Alternative 1 to address the issue raised at L324 (apologies in advance if it's *too* odd), but could the following two examples be equivalent?

```
// ...
{
    use q = Qubit();
    // ...
}
// ...
```

```
// ...
use q = Qubit() in {
    // ...
}
// ..
```

That is, would it be possible to make the distinction between a statement and a block the use of the `in` keyword to explicitly denote a scope? Doing so would possibly also allow for more functional-style `let` blocks, perhaps?

```
let foo = bar in {
    // foo has the value bar here.
}
// foo is unbound here.
```

cgranade (commented):
I love this, thanks for tagging me in! I had some fairly general thoughts and some musings about possible other directions to explore, but on the whole I think this could be a significant improvement in the ergonomics of using Q#.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+Example 2a: Many qubits allocated in a single `using` statement with tuple destructuring.
+
+```qsharp
+operation QubitTuple(n : Int) : Result {
+    using ((a, b, c) = (Qubit(), Qubit[2 * n + 1], Qubit[n])) {
+        // ...
+        return M(q);
```
ScottCarda-MS, line 71: Since `q` is not one of the qubits allocated in the example, it is a bit confusing to include it here. It is possible the qubit was allocated somewhere else, but it may be simpler to replace this reference to refer to one of the qubits that are part of the example, like `a`.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+
+Example 3a: Many qubits allocated with a separate `using` block for each variable.
+
+```qsharp
+operation NestedBlocks(n : Int) : Result {
+    using (a = Qubit()) {
+        using (b = Qubit[2 * n + 1]) {
+            using (c = Qubit[n]) {
+                // ...
+                return M(q);
```
ScottCarda-MS, line 84: same here

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+
+Example 3b: Many qubits allocated with a separate `use` statement for each variable.
+
+```qsharp
+operation NestedBlocks(n : Int) : Result {
+    use a = Qubit();
+    use b = Qubit[2 * n + 1];
+    use c = Qubit[n];
+    // ...
+    return M(q); // a, b, and c are released here.
```
ScottCarda-MS, line 217: same here

Approved/implicitly-scoped-qubit-allocation.md
ScottCarda-MS: I hadn't seen that proposal before, but as I read through this I could help but think of this approach as well. It doesn't seem like there is really much of a distinction between `use q = Qubit()` and `let q = Qubit()`. The only thing you really would want the `use` for is to distinguish between whether or not you are getting a "clean" qubit or not, i.e. whether or not you are borrowing the qubit. To that end the `use` and `borrow` sort of take the place of the `new` keyword from other languages: `let q = use Qubit();` `let w = borrow Qubit();`.

ScottCarda-MS: And actually at that point we may as well just use the `new` keyword instead of `use`, as it is not only familiar to developers but it makes more sense in this context purely from reading the statement out loud. The `borrow` keyword here would still be useful and makes sense: `let q = new Qubit();` and `let w = borrow Qubit();`.

samarsha: My issue with putting `use` on the right is that we don't want to allow this:

```qsharp
use a = use Qubit();
mutable b = use Qubit();
set b = a; // Is the original qubit of b released here?
```

or this:

```qsharp
Foo(use Qubit()); // When is the qubit released? The scope is unclear.
```

That means we need more complicated rules:

1. Qubit initializers can only appear in a `use` expression.
2. `use` expressions can only appear on the right-hand side of a `let` statement.

Compare to: qubit initializers can only appear on the right-hand side of a `use` statement.

About `new Qubit()`, I think that Q# tries to avoid the connotation that qubits are OOP-style objects. Operations on an allocated qubit can implicitly affect the entire quantum state, so they aren't new, independent objects like the `new` keyword would imply.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: I actually wouldn't consider this bad style, necessarily - assuming `q` is actually used for something after it's allocated. :) Perhaps you want to do something with `foo` but limit its use to that scope. I think it's OK to use the scope statement for scoping of both classical and quantum variables.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: Yes, it's a breaking change. I'm not sure if we want to go through a keyword opt-in while we are pre-1.0. We should probably decide how we want to handle new keywords in general.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: I think that's fine. It's allowed by the current `using` statement and is consistent with `let`.

Approved/implicitly-scoped-qubit-allocation.md
ScottCarda-MS: I'm probably missing some context here, but why wouldn't they always be completely independent at the moment of the assignment statement? Talking specifically about `new`/`use` and not `borrow`.

ScottCarda-MS: We could simply disallow mutable variables holding qubits. Is there something about using a `let` that means we also have to allow `mutable` for the same situation? If so, that would really be the only difference between a `let` and a left-side `use`: the connotation that the `let` could be replaced for a `mutable`. But I don't expect that that connotation is especially true or that we couldn't simple disallow it in this case.

samarsha: > I'm probably missing some context here, but why wouldn't they always be completely independent at the moment of the assignment statement?

Phase kickback, for example:

```qsharp
@EntryPoint()
operation Main() : Unit {
    using (a = Qubit()) {
        FlipPhase();
        DumpMachine();
    }
}

operation FlipPhase() : Unit {
    using (b = Qubit()) {
        within {
            X(b);
        } apply {
            Z(b);
        }
    }
}
```

`FlipPhase()` changes `a` from |0> to -|0> by only applying operations to `b`.

samarsha: About `use` on the right, I think my argument wasn't quite right - if `use Qubit()` is one unit, then the rule is that it can only appear on the right of a `let` statement, similar to how `Qubit()` can only appear on the right of a `use` statement, so that is not too bad. The other issue I have though, is that the lifetime of a qubit is tied to the scope of a variable - `use q = Qubit()` makes it clear that when `q` goes out of scope, the qubit is released. `let q = use Qubit()` is not as clear what the lifetime is IMO - `q` looks like a normal, non-resource variable.

For comparison, F# also does `use x = foo ()`, not `let x = use foo ()`, and C# does `using var x = Foo();`, not `var x = using Foo();`.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: Yes, that sounds like a good idea.

Approved/implicitly-scoped-qubit-allocation.md
ScottCarda-MS: Is the only difference then that you can't have a return statement with a variable declared with `use`?

ScottCarda-MS: I guess I'm trying to see the fundamental difference in logic between the `use` and the `let` statements.

Approved/implicitly-scoped-qubit-allocation.md
cgranade: Makes sense, just wanted to ask for clarification. Thanks!

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,342 @@
...
+title: Implicitly-scoped qubit allocation
+description: Provides more convenient syntax for allocating qubits.
+author: Sarah Marshall
+date: September 28, 2020
+---
+
+# Proposal
+
+1. Add new `use` and `borrow` statements to allocate qubits that are valid until the end of the current block.
+2. Add new scope statement that does not require a preceding keyword, and is executed exactly once; equivalent to `if (true) { }`.
```
ScottCarda-MS, line 11: I take it this is useful in this proposal for terminating the lifetime of an allocated qubit. Does it have uses beyond that?

samarsha: It limits the scope of any variable, including classical variables declared with `let` or `mutable`, see [Impact on Existing Mechanisms](https://github.com/microsoft/qsharp-language/blob/samarsha/qubit-allocation/Approved/implicitly-scoped-qubit-allocation.md#impact-on-existing-mechanisms).

Approved/implicitly-scoped-qubit-allocation.md
samarsha: The difference is that the value of variables declared with `use` become invalid outside the scope, while the value of variables declared with `let` don't.

```qsharp
mutable numbers = EmptyArray<Int>();
{
    let xs = [1, 2, 3];
    set numbers = xs;
}
Message($"{numbers[2]}");
```

works OK because while `xs` is out of scope, the array it pointed to is still valid. But

```qsharp
mutable qubits = EmptyArray<Qubit>();
{
    use qs = Qubit[3];
    set qubits = qs;
}
X(qubits[2]);
Message($"{MResetZ(qubits[2])}");
```

doesn't work; even though we tried to save the qubit array, the qubits in the array are no longer valid because `qs` went out of scope.

alan-geller: It's pretty much the same as F# or C# -- something you "create" with `use` is `IDisposable` and gets disposed of (the `Dispose` method gets called) when it goes out of scope. We don't have interfaces, but qubits are the Q# equivalent of disposable.
In all 3 languages, the `use` keyword on the left indicates that there's special processing of the *value* (not the variable!) related to when the variable goes out of scope.

ScottCarda-MS: So then the `use` is really more attached to the value of the qubit(s) rather than the container, doesn't it make more sense then to have it closer to the value as in `let q = use Qubit();` rather than closer to the container in `use q = Qubit();`? Also the differentiating scenario you posted is prevented when we prevent mutables from holding qubit references, as previously suggested. To that end, we could look at a scenario where qubit references are returned from a callable:
```
operation Foo() : Qubit {
    using (q = Qubit()) { return q; }
    // or
    use q = Qubit();
    return q;
    // or
    let q = use Qubit();
    return q;
}
```
I believe this scenario is currently allowed, which to me begs the question of what is supposed to happens when this operation is called, but this could also be prevented by not allowing the Qubit type to be anywhere in the return type of a callable. Of course the other approach would be to be more permissive with the Qubit type and allow its scoping rules and lifetime to be the same as other types.

Approved/implicitly-scoped-qubit-allocation.md
alan-geller: F# used to work this way; let statements explicitly introduced a new block. They made this optional in 2.0, I think, and then dropped it entirely later on. I think the "bound until end of the containing scope" style is much easier to write and read, partly because it leads to less indentation -- less visual noise.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: > So then the `use` is really more attached to the value of the qubit(s) rather than the container

Well, it's both. When the container goes out of scope, the value it contains invalidated. That's why I think it's more important to emphasize the container than the value, because the container decides what happens to the value and when.

> I believe this scenario is currently allowed, which to me begs the question of what is supposed to happens when this operation is called

It's undefined. The qubit is not valid after it's returned. Yes, ideally the compiler would prevent this.

> this could also be prevented by not allowing the Qubit type to be anywhere in the return type of a callable

That is too strict. This is a well-defined function:

```qsharp
function FlipPair(q1 : Qubit, q2 : Qubit) : (Qubit, Qubit) {
    return (q2, q1);
}
```

> the other approach would be to be more permissive with the Qubit type and allow its scoping rules and lifetime to be the same as other types

Qubit lifetimes being scope-limited is a deliberate design choice made by Q#. I think it's unlikely to change.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: F# still works this way, actually (but it is optional):

```fsharp
let x = 10 in
    printfn "x is %i" x
```

To be more precise, the full form is `let x = v in e`, where `e` is an expression (not a block). My understanding is that

```fsharp
let x = 10
printfn "x is %i" x
```

is actually just syntactic sugar for the first version. The first version is still useful sometimes when you want to define a variable and use it on the same line, for example:

```fsharp
[ 1; 2; 3 ]
|> List.map (fun x -> let y = 2 * x in y * y)
```

samarsha: And playing around with this a little more, I realized that

```fsharp
let x = 10 in
    printfn "x is %i" x
printfn "x is %i" x
```

actually prints "x is 10" twice. I think this is because the `in` absorbs all the remaining expressions even if they don't match the initial indentation. This might be to support OCaml-style nested let bindings:

```fsharp
let x = 10 in
let y = x + 1 in
printfn "%i %i" x y
```

So it's not exactly the same as @cgranade's suggestion, but adding a `do` makes it indentation-sensitive, and `x` is undefined on the last line:

```fsharp
do let x = 10 in
    printfn "x is %i" x
printfn "x is %i" x
```

Approved/implicitly-scoped-qubit-allocation.md
ScottCarda-MS: Okay I think I'm getting it. So to summarize the difference
- `let` declares a container holding a value that may outlive the container and the current scope.
- `mutate` declares a container holding a value where the value may outlive the container, and the contain may outlive the value (for example if it is assigned another value).
- `use` declares a container holding a value that will end (is there a better way to phrase this?) when the container ends, which will be when the current scope ends.
- `borrow` is the same as `use` except the internal state of the value may be different.

Does that all seem correct?

ScottCarda-MS: Oh would we then disallow:
```
use q = Qubit();
{
   use w = q;
}
```
In this case the value in the container `w` would be invalidated after the end of the inner scope, which would invalidate the value of `q` before `q` goes out of scope.

samarsha: The way I think about it is:

* `let` gives a name to a value. The name always refers to the same value.
* `mutable` gives a name to a value. The name can refer to different values at different times.
* `use` / `borrow` gives a name to a value. When the name goes out of scope, something like `Release(q)` is called for all qubits `q` contained in the value.

> Oh would we then disallow:
>
> ```qsharp
> use q = Qubit();
> {
>    use w = q;
> }
> ```

Because qubit initializer expressions can only contain `Qubit()`, `Qubit[n]`, and tuples of those, this is not allowed currently and is not allowed by this proposal, either.

cgranade: > It's pretty much the same as F# or C# -- something you "create" with `use` is `IDisposable` and gets disposed of (the `Dispose` method gets called) when it goes out of scope. We don't have interfaces, but qubits are the Q# equivalent of disposable.
> In all 3 languages, the `use` keyword on the left indicates that there's special processing of the _value_ (not the variable!) related to when the variable goes out of scope.

Two points by way of agreement:

- Python has very similar in the form of a `with` block:
  ```python
  with open('a.txt', 'r') as f:
      contents = f.read()
  # f and contents are both valid here, but the with block
  # ensures that f.__exit__ has been called, even in the case
  # of exceptional flow control.
  ```

- C# has recently introduced `using var` as a new kind of declaration that acts very similarly to the new `use` proposed here:
  ```csharp
  {
      // ...
      using var z = new SomethingDisposable();
      // ...
  } // z.Dispose will get called here, even with exceptional flow control.
  ```

Approved/implicitly-scoped-qubit-allocation.md
cgranade: > F# still sort of works this way (but it is optional):

That's kind of what I was going towards. Having something that works well for low visual noise like statement-level declarations is a great idea, but sometimes (like in the phase kickback example) it can be really nice to be very explicit about the scope of a qubit in a visually obvious way.

```Q#
use flag = Qubit() in {
    within {
        H(flag);
        X(flag);
    } apply {
        oracle();
    }
}
```

samarsha: I think my preference is to compose `{ }` and `use` as separate statements that can be used together or not, rather than having two variants of `use` (one ending with `in { }` and one ending with `;`). I do prefer `let ... in ...` as an expression that takes a binding and another expression, but it feels weird to me as a statement that takes a binding and a block. The `in` version makes really makes me wish I could do

```qsharp
Foo(let x = 10 in x * x);
```

but that wouldn't be allowed in the statement version.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: It's good to think about. I'm hoping it's mostly orthogonal to this proposal, since changes to qubit initializer expressions shouldn't need to affect the overall `use` statement. However, as you mentioned if we change initializers to put `use` on the right-hand side, like `let qs = LittleEndian(use Qubit[10]);` then that would affect this proposal.

As discussed in the other thread, I'm in favor of keeping `use` on the left, even if initializers are expanded to contain non-`Qubit` expressions like `LittleEndian(Qubit[10]);`. I think it'd be good to come to consensus on the positioning of `use` and then we could have a separate proposal for adding UDT wrapping and state preparation to qubit initializers?

cgranade: I totally agree with `use` on the left; my only point in raising was that `use` on the right has come up enough it's helpful to document as a considered and rejected alternative. In particular, that `use` on the right makes the state preparation / UDT discussion harder is a great argument in favor of `use` on the left.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: Added a note about this to Further Considerations.

Approved/implicitly-scoped-qubit-allocation.md
bettinaheim: Hm, I am not sure if the change in keyword makes sense. What are the arguments for it?

samarsha: Generally, I think keywords should use simple word forms. We say `return x;`, not `returning x;`, `open Foo;`, not `opening Foo;`, etc. There are no other keywords in Q# that end in -ing besides `using` and `borrowing`, so we will be gaining consistency with all other keywords, not just `let`.

In terms of consistency with other languages like C#, it does not seem like there is any consensus among other programming languages. While C# has `using`, F# has `use` instead. Java re-uses the `try` keyword, and Python has `with`. I think similarity with C# is not a strong enough argument to go with `using`.

Approved/implicitly-scoped-qubit-allocation.md
bettinaheim: Agreed, this would be very good to look into. We should also examine the exact syntax of the initializers in general - Qubit[n] seems like the odd one out if we want to remove new array expressions (which I think is a good thing).

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,347 @@
...
+}
+```
+
+## Comparison to Alternatives
+
+### Alternative 1
+
+This alternative has the advantage of being backwards-compatible, but introduces inconsistencies and added complexity:
+
+1. The `using` and `borrowing` keywords are re-used in different contexts: sometimes it is a block statement and sometimes it is a non-block statement.
```
bettinaheim, line 329: Since in both cases the intent of the keywords remains the same, I don't see that as problematic necessarily. 
The fixup block in repeat loops for example is optional, such that both "... until(condition);" and "...until(condition) fixup {...}" may occur.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,347 @@
...
+   New keywords `use` and `borrow` could be introduced for the non-block statements, but this adds redundancy and keyword clutter.
+2. The block statement requires parentheses around the declaration, but the non-block statement does not.
```
bettinaheim, line 331: I agree that dropping the parens would be good if we keep the block version. In that case, we should also drop the requirement for parenthesis after keywords for other block statements (if-, repeat-, while-, and for-statement), and list it under impact on existing mechanisms.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,347 @@
...
+title: Implicitly-scoped qubit allocation
+description: Provides more convenient syntax for allocating qubits.
+author: Sarah Marshall
+date: September 28, 2020
+---
+
+# Proposal
+
+1. Add new `use` and `borrow` statements to allocate qubits that are valid until the end of the current block.
+2. Add new scope statement that does not require a preceding keyword, and is executed exactly once; equivalent to `if (true) { }`.
```
bettinaheim, line 11: I am a bit skeptical about the deprecation of the block-version in favor of a scope statement (see also my comments below).

Approved/implicitly-scoped-qubit-allocation.md
bettinaheim: It would be good to summarize briefly the points on this thread either under the related mechanism section, to which we should add considerations around initializers in general.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: To flip the question: what reasons are there to name it `using` instead of `use`?

bettinaheim:
@samarsha I think I have some decent ideas on the initializers, and I'll write them up shortly.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: The scope statement has been removed from the proposal, so this syntax is no longer valid.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,347 @@
...
+        H(q);
+        return M(q);
+    }
+}
+```
+
+In combination with the new scope statement, the lifetime control provided by `using` can be obtained:
+
+```qsharp
+operation FlipCoin() : Result {
```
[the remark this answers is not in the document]
samarsha: The scope statement has been removed, and `use` must precede the block now.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,347 @@
...
+
+The new keywords `use` and `borrow` were previously valid Q# identifiers.
+Adding them is a breaking change, unless an opt-in mechanism for new keywords is added to the language.
+
+The scope statement added in this proposal can also be used to limit the scope of non-qubit variable bindings declared with `let` and `mutable`.
+This may be useful for limiting the lifetime of large data structures like arrays or for otherwise communicating the intent that a variable should only be used in a specific section within a callable.
+
+## Anticipated Interactions with Future Modifications
+
+Because the proposed modification is similar to existing functionality, it is not expected to have any impact on future language development.
```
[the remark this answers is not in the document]
samarsha: The scope statement has been removed.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: The `use` and `borrow` statements support optional blocks now.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: The proposal removes parentheses from all block statement headers now.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: The scope statement has been removed, and `use` and `borrow` take optional blocks.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: @bettinaheim Based on discussions offline, it sounds like you are OK with `use` and `borrow` instead of `using` and `borrowing`. I think we should still decide how to handle the new keywords: should they immediately break code that has identifiers named `use` and `borrow`, should we have an opt-in mechanism, or should we deprecate identifiers named `use` and `borrow` for a period of time?

Approved/implicitly-scoped-qubit-allocation.md
alan-geller: At this point, the block form of `use` is pretty much the same as the current `using` statement. Maybe we should we just keep `using` for this form and reserve `use` for the statement form?

cgranade: @samarsha: Agreed, the opening brace is definitely clear enough to denote a new scope. I like that the new form also requires one less pair of parentheses, making it a lot cleaner to read. As for `use` versus `using`, I think `use` and `borrow` have a nice parity with `let` and `set`, such that my suggestion would be to adopt `use` and `borrow`. For backwards compatibility, we could maybe keep `using` and `borrowing` as deprecated keywords?

Approved/implicitly-scoped-qubit-allocation.md
bettinaheim: To quickly just recapitulate: use and borrow could be more convenient to use and there is no strong case based on what is used in other languages. I am in generally worried about unnecessary changes, and I wouldn't want to consider "keyword tweaks" past 1.0. Given that we are still in preview, we can consider this one. @viveis for input as well.

Approved/implicitly-scoped-qubit-allocation.md
viktorveis: Could someone please help me understand why do we need to keep the block instead of the original proposal?
```qsharp
use q = Qubit();
```

cgranade: @viktorveis: Both forms are quite useful in their own respects, I think. For the case in which a qubit allocation scope is the same as an operation scope, the statement form eliminates a lot of unneeded levels of nesting. That said, whenever you're doing something like phase kickback, where a temporary qubit is only needed for a short time, retaining a block form of `use` and `borrow` can help make sure that the temporary qubit is released as soon as possible and can make visually clear the scope in which the temporary qubit is used.

samarsha: > Maybe we should we just keep `using` for this form and reserve `use` for the statement form?

I would like to be more economical in keyword usage than this, since I think needing to remember whether to write `use` or `using`, or `borrow` or `borrowing`, would be an unnecessary burden on users, when the compiler is capable of figuring out whether you meant the block or non-block form based only on whether a block follows the statement or not.

While Q# has introduced very similar keywords in the past - like `Controlled` vs `Ctl` vs `controlled`, I am not so sure that's a good idea, because personally I find having to remember which spelling of "controlled" I need in this particular context is kind of annoying.

ScottCarda-MS: Does having this syntax:
```
use q = Qubit() {
    // ...
}
```
mean that we will support the following syntax in the future?
```
let a, b, c = 12, "Yes", 3.0 {
    // ...
}
```
where `a`, `b`, and `c` are scoped to to following block? Seems like we already have a syntax similar to that with the specialization declaration syntax:
```
operation Foo(a : Int, b : String, c : Double) : Unit is Ctl {
    body(...) {
        // The contents of the parentheses define all the bindings at the start of this scope.
        // The "..." bring in all the super-scope's bindings.
    }
    controlled (control, ...) {
        // Here, in addition to the super-scope's bindings, an additional binding for the control qubits is included.
    }
}

Comment (ScottCarda-MS) [reply Approved/implicitly-scoped-qubit-allocation.md:259]: If this could be adapted to a more general form that would allow assignment, it could be used in general to define callable definitions, specialization definitions, qubit lifetimes, and any arbitrary scopes that a coder may want to initialize, possibly for the purpose of shadowing variables. It could really simplify the construction of the language. Is this of interest?

Comment (samarsha) [reply Approved/implicitly-scoped-qubit-allocation.md:259]: > Does having this syntax:
> 
> ```
> use q = Qubit() {
>     // ...
> }
> ```
> 
> mean that we will support the following syntax in the future?
> 
> ```
> let a, b, c = 12, "Yes", 3.0 {
>     // ...
> }
> ```

There would be consistency from that, but I don't think having a block `use` statement necessarily means we will or should support a block `let` statement.

> If this could be adapted to a more general form that would allow assignment, it could be used in general to define callable definitions, specialization definitions, qubit lifetimes, and any arbitrary scopes that a coder may want to initialize, possibly for the purpose of shadowing variables.

I am not sure what the more general form you're referring to is, can you clarify?

File: Approved/implicitly-scoped-qubit-allocation.md [hunk 18]
```diff
@@ -0,0 +1,291 @@
...
+title: Implicitly-scoped qubit allocation
+description: Provides more convenient syntax for allocating qubits.
+author: Sarah Marshall
+date: TBD
+---
+
+# Proposal
+
+1. Rename `using` and `borrowing` keywords to `use` and `borrow`.
+2. Allow `use` and `borrow` statements to be followed by either a block `{ ... }` or a semicolon `;`.
```
samarsha, line 11: The more I think about this, the more I would like to go back to the original proposal with the scope statement and without the block `use` statement. One of the primary goals of this proposal is to avoid forcing users to decide between excessive block nesting and unnecessary tuple destructuring. But by removing the scope statement, and having a block `use` statement instead, we continue to have that exact problem whenever you want to allocate more than one qubit variable in a short-lived scope:

```qsharp
X(q1);
use (q2, q3) = (Qubit(), Qubit()) {
    // ...
}
return M(q1);
```

I would prefer instead:

```qsharp
X(q1);
{
    use q2 = Qubit();
    use q3 = Qubit();
    // ...
}
return M(q1);
```

@bettinaheim @cgranade @viktorveis

Approved/implicitly-scoped-qubit-allocation.md
ScottCarda-MS: I wasn't referring to something specific, just an idea to consider, if we wanted to go down that route. However the scope syntax being suggested in this PR is pretty general and could be used in a variety of places, so it may cover what I'm suggesting here. I like the idea of having the generic scope syntax being suggested over the `use` blocks as I think it will provide more consistency across the language's syntax.

Approved/implicitly-scoped-qubit-allocation.md
ScottCarda-MS: I like the scope syntax more than the `use` blocks as I think it is more syntactically consistent with what we already have the in language. The `use q = Qubit();` is consistent with how we define other symbols. And the scope syntax could almost be seen as a generalization of how we declare scopes for other language constructs like callables and for loops. This syntax could be built upon to possible define lambdas, if we choose to go down that route.

Approved/implicitly-scoped-qubit-allocation.md
alan-geller: Having come this far -- which I think is quite reasonable -- is there a reason we don't continue with `using` for the block case and `use` for the statement case? This would reduce the amount of churn in user and library code at the cost of two additional keywords.
For what it's worth, there's a reasonable "English" reason for `using` being present progressive -- something that's true during a process -- and `use` being a command, "do it now"...
If we ever want block-form `let`, I think this would argue for `letting` being the keyword, which actually doesn't feel wrong to me. `Mutating` is a little bit more odd -- although it's not clear to me how much value there is in such a construct in actual use.

All that said, I'm also fine with having `use` be the common keyword between the two forms.

samarsha: @alan-geller Possibly my reply was lost since this is a long thread, but here was my reply to having both `use` and `using`:

> > Maybe we should we just keep `using` for this form and reserve `use` for the statement form?
> 
> I would like to be more economical in keyword usage than this, since I think needing to remember whether to write `use` or `using`, or `borrow` or `borrowing`, would be an unnecessary burden on users, when the compiler is capable of figuring out whether you meant the block or non-block form based only on whether a block follows the statement or not.
> 
> While Q# has introduced very similar keywords in the past - like `Controlled` vs `Ctl` vs `controlled`, I am not so sure that's a good idea, because personally I find that having to remember which spelling of "controlled" I need in a particular context is confusing.

samarsha: Regarding churn in existing Q# code, my primary concern while we are pre-1.0 is to make sure we have a cohesive design (e.g. without redundant keywords). I view backwards compatibility as a secondary concern to that. Once we reach 1.0, and no longer have the freedom to make minor adjustments like this, those two priorities will switch places IMO.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,347 @@
...
+## Timeline
+
+There are no dependencies on other proposals or libraries.
+The work needed for either option appears to be minimal, since it is similar to the existing implementation.
+
+# Further Considerations
+
+Development tools should consider adding support for transitioning from the current syntax to the proposed syntax, such as a code action in IDEs.
+
+## Related Mechanisms
```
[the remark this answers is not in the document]
samarsha: I think I added notes about everything discussed, please let me know if I missed something.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: After discussing this with @bettinaheim offline, we agreed that it was better to avoid making major changes to the existing block statement syntax until we know more about how the non-block syntax will be used in practice, and if there is actually a need to better support more advanced allocation with an explicit scope.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+title: Implicitly-scoped qubit allocation
+description: Provides more convenient syntax for allocating qubits.
+author: Sarah Marshall
+date: TBD
+---
+
+# Proposal
+
+1. Rename `using` and `borrowing` keywords to `use` and `borrow`.
+2. Allow `use` and `borrow` statements to be followed by either a block `{ ... }` or a semicolon `;`.
```
bettinaheim, line 11: It would be good to document that we defer the call on whether to permanently support the block version until we have further insights into how allocations are used with this new option now being available.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+1. It requires creating a new block to define the lifetime of the qubit, even when the intended lifetime is the same as the current block.
+2. Allocating multiple qubits at the same time requires either tuple destructuring or nested blocks, both of which can be hard to read.
+
+As a quantum programming language, qubit allocation is at the core of Q#.
+Both the simple case of allocating a single qubit for the remainder of the current block, and more complicated scenarios involving multiple qubit variables that are both clean and borrowed and have varying lifetimes, should have clear, intuitive syntax that doesn't require boilerplate.
+
+# Description
+
+Qubits are a managed resource in Q#: when qubits are allocated, they are valid only within the new block declared at allocation, after which they are released and can no longer be used.
+This proposal makes qubit allocation in Q# always scoped to the current block.
```
bettinaheim, line 27: I believe always is incorrect here; the proposal introduces the option to allocate qubits that are scoped to the current block.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+Instead of being released at the end of a newly-declared block, qubits are released at the end of the current block.
+It also extends block statement syntax to allow precise control of qubit lifetime when needed in more complicated scenarios.
```
bettinaheim, line 29: I think this sentence also needs updating.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+This syntax is preserved by this proposal with only minor changes.
+
+## Proposed Modification
+
+The new `use` and `borrow` statements behave like the existing `using` and `borrowing` statements with two differences:
+
+1. The block is optional.
+   If no block is provided, the scope of the qubit is implicitly defined to be from the `use` or `borrow` statement until the end of the current block.
+2. Parentheses around the header are not required.
+   For consistency, parentheses are no longer required around any block statement header.
```
bettinaheim, line 125: Could you please explicitly list all statements for which there will be a change? Could you please also clarify whether the intent is to have the parenthesis be permanently optional or whether the intent is to ultimately error on parenthesis that are currently valid?

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+description: Provides more convenient syntax for allocating qubits.
+author: Sarah Marshall
+date: TBD
+---
+
+# Proposal
+
+1. Rename `using` and `borrowing` keywords to `use` and `borrow`.
+2. Allow `use` and `borrow` statements to be followed by either a block `{ ... }` or a semicolon `;`.
+3. Remove the requirement for parentheses around headers for all block statements.
```
bettinaheim, line 12: Is the intent to make the parenthesis permanently optional, or to ultimately deprecate their use (i.e. error on use)?

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+# Implementation
+
+The `use` and `borrow` keywords can be added, and the `using` and `borrowing` keywords can be deprecated.
+Until `using` and `borrowing` are removed, they can be used in place of `use` and `borrow` for both the block and non-block statements.
+
+The block for the existing `using` and `borrowing` syntax nodes can be made optional.
+Either the block remains optional throughout all stages of the syntax tree, or the non-block form is transformed into the block form by automatically moving all statements below it into a new block.
+In the former case, all syntax tree consumers need to be aware of both forms, while in the latter case consumers would notice no change from the current syntax tree node.
```
bettinaheim, line 196: I'd rather go with the latter. Do you see any issue with that?

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+## Anticipated Interactions with Future Modifications
+
+There will likely be changes to qubit initializers soon.
+See [Issue #40: Allocatable types and generalization of initializers](https://github.com/microsoft/qsharp-language/issues/40).
+We expect the changes to initializers to integrate well with this proposal, because this proposal only changes the syntax *around* qubit initializers, not the initializers themselves.
+
+Removing parentheses around block statement headers may affect future syntax development.
+For example, it would not be possible to make the braces optional without also re-introducing parentheses or another token, because of parsing issues with code like:
```
bettinaheim, line 237: The better alternative to mention in this case I believe would be to move to having a keyword taking the role of the delimiter, e.g. 
```
if M(q) == One then
    X(q);
```

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+In the former case, all syntax tree consumers need to be aware of both forms, while in the latter case consumers would notice no change from the current syntax tree node.
+
+## Timeline
+
+There are no dependencies on other proposals or libraries.
+The work needed for either option appears to be minimal, since it is similar to the existing implementation.
+
+# Further Considerations
+
+Development tools should consider adding support for transitioning from the current syntax to the proposed syntax, such as a code action in IDEs.
```
bettinaheim, line 205: We should also gather telemetry to assess how often people choose to go with the block version. Let's discuss offline what data we need to gather to inform a possible future evolution for the block version.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,314 @@
...
+
+Qubit allocation is currently awkward in Q# for two reasons:
+
+1. It requires creating a new block to define the lifetime of the qubit, even when the intended lifetime is the same as the current block.
+2. Allocating multiple qubits at the same time requires either tuple destructuring or nested blocks, both of which can be hard to read.
+
+As a quantum programming language, qubit allocation is at the core of Q#.
+Both the simple case of allocating a single qubit for the remainder of the current block, and more complicated scenarios involving multiple qubit variables that are both clean and borrowed and have varying lifetimes, should have clear, intuitive syntax that doesn't require boilerplate.
+
+# Description
```
bettinaheim, line 24: It won't hurt to mention explicitly either here or under implementation that the compiler may deallocate qubits before the allocation scope ends if they are no longer in use.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: The intent is to make parentheses a syntax error, except in cases where the parentheses contain a single expression, in which case it will continue to be supported since any expression can be wrapped in parentheses.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: No, I think that could work, but I left it open-ended here in case things change during implementation.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,323 @@
...
+description: Provides more convenient syntax for allocating qubits.
+author: Sarah Marshall
+date: October 27, 2020
+---
+
+# Proposal
+
+1. Deprecate `using` and `borrowing` keywords and replace with `use` and `borrow`.
+2. Allow `use` and `borrow` statements to be followed by either a block `{ ... }` or a semicolon `;`.
+3. Deprecate and remove parentheses around block statement headers.
```
ScottCarda-MS, line 12: Is this something that was decided on based on user data? I personally find the parentheses helpful in quickly parsing the code while looking through it.

Approved/implicitly-scoped-qubit-allocation.md
ScottCarda-MS: I would prefer to have the parentheses optional at least as it makes the code easier to read for people like me coming from a C++/C# background. Was this decided based on user data?

samarsha: Could you clarify how

```qsharp
for (x in xs) {
   // ...
}
```

is more readable than

```qsharp
for x in xs {
    // ...
}
```

?

samarsha: About making them optional, I would prefer not to have too many ways to express the same construct in Q#, so I think we should pick one or the other rather than supporting both. If the non-parenthesized form has an issue with readability, then I don't think making it optional will fully solve that problem because code written by other people could still choose not to use parentheses.

ScottCarda-MS: If we want to go one way versus the other permanently on account of readability, then we should make a decision on what is more readable for our users based on user data, rather than just our perspectives. I'm tempted to defend the readability of parentheses here in an answer to your question above, but I think that would undermine my point of using user data 😊.

samarsha: The parentheses were removed for consistency with the newly added non-block `use` and `borrow` statements. Since we now have

```qsharp
use q = Qubit();
```

It's consistent for the block form to be

```qsharp
use q = Qubit() {
    // ...
}
```

instead of

```qsharp
use (q = Qubit()) {
    // ...
}
```

We would need an argument that removing the parentheses has some cost that is greater than the benefit gained from consistent syntax. I don't think user data by itself is able to provide that argument.

Approved/implicitly-scoped-qubit-allocation.md
samarsha: I'd like to keep this conversation in the other thread that is also about removing parentheses.

Approved/implicitly-scoped-qubit-allocation.md
bettinaheim: Since I don't see any significant difference in readability, I am fine with making this call here. Allowing the parenthesis to me doesn't add enough value to justify the existence of a redundant way of expressing the same concept (see also our [guidelines](https://github.com/microsoft/qsharp-language/blob/main/Guidelines.md)). For consistency reasons I agree with dropping the parentheses.

Approved/implicitly-scoped-qubit-allocation.md
bettinaheim: In general, if we introduce new items into the syntax tree (without a compiler step to eliminate it if desired), then we force every extension and backend that deals with the syntax tree to do work. Hence it is good to have clarity around how much work we impose on whom. :) I would clarify that here. 
In terms of process, it is of course possible that we run into unexpected issues with how we envisioned the implementation. That why we open an issue to track the status of the implementation when it starts and to discuss precisely if and how the proposal needs to be adapted. Having it stated explicitly here hence doesn't mean that we can't revise it during implementation, but it means there should be a good reason to revise it.

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,322 @@
...
+
+We expect to permanently support the new non-block `use` and `borrow` statements, but we may reconsider how or if the block forms are supported at a later date, after we learn more about how both forms are used in practice.
+
+The parentheses that are currently required by the `for`, `using` or `use`, and `borrowing` or `borrow` statements will be deprecated and removed.
+Including parentheses with these statements will become a syntax error.
```
bettinaheim, line 17:
```suggestion
Including parentheses with these statements will become a syntax error with the next major version.
```

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,322 @@
...
+
+This proposal adds the option to allocate qubits that are scoped to the current block, instead of a new block.
+Its aim is to be syntactically concise and consistent with other binding statements like `let`.
+As a consequence, it also modifies the existing block qubit allocation syntax slightly in order to increase consistency with the new non-block form.
+
+## Current Status
+
+Currently, the only mechanism to allocate qubits in Q# is with `using` and `borrowing` block statements.
+No new functionality is provided by this proposal.
+Every example with the existing syntax has a corresponding equivalent example with the proposed syntax, and vice versa.
```
bettinaheim, line 44: I'd just remove that sentence, since it is not perfectly correct (in both directions), though I agree with the sentiment.
```suggestion
```

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,322 @@
...
+This example shows the main benefit of the current syntax: it allows the lifetime of a qubit to be shorter than the parent block.
+This syntax is preserved by this proposal with only minor changes.
+
+## Proposed Modification
+
+The new `use` and `borrow` statements behave like the existing `using` and `borrowing` statements with two differences:
+
+1. The block is optional.
+   If no block is provided, the scope of the qubit is implicitly defined to be from the `use` or `borrow` statement until the end of the current block.
+2. Parentheses around the header are not required.
```
bettinaheim, line 132:
```suggestion
2. Parentheses around the header are not required and will be no longer supported after then next major version.
```

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,322 @@
...
+   For consistency, parentheses are no longer required around any block statement header.
```
bettinaheim, line 133:
```suggestion
   For consistency, the same holds for parentheses around any block statement header.
```

Approved/implicitly-scoped-qubit-allocation.md
```diff
@@ -0,0 +1,322 @@
...
+    }
+    // Classical post-processing...
+    return r;
+}
+```
+
+# Implementation
+
+The `use` and `borrow` keywords can be added, and the `using` and `borrowing` keywords can be deprecated.
+Until `using` and `borrowing` are removed, they can be used in place of `use` and `borrow` for both the block and non-block statements.
```
bettinaheim, line 200:
```suggestion
Until `using` and `borrowing` are removed, they can be used in place of `use` and `borrow` for both the block and non-block statements, but their use will generate a warning.
```

bettinaheim (approved):
Thanks for working this out!

PR #34, opened 2020-09-29
Value restrictions for exponentiation

bettinaheim:
Documenting that the behavior of exponentiation is undefined if the exponent is smaller than zero or larger than the 32-bit integer maximum.
Also, fixing some spelling mistakes.


Specifications/Language/3_Expressions/ArithmeticExpressions.md
```diff
@@ -1,7 +1,11 @@
 # Arithmetic Expressions
 
 Arithmetic operators are addition (`+`), subtraction (`-`), multiplication (`*`), division (`/`), negation (`-`), exponentiation (`^`). 
-They can be applied to operands of type `Int`, `BigInt`, or `Double`. Additionally, for integral types (`Int` and `BigInt`) an operator computing the modulus (`%`) is available. For binary operators, the type of both operands has to match, expect for exponentiation; an exponent for a value of type `BigInt` always has to be of type `Int`. The type of the entire expression matches the type of the left operand. 
+They can be applied to operands of type `Int`, `BigInt`, or `Double`. Additionally, for integral types (`Int` and `BigInt`) an operator computing the modulus (`%`) is available. 
+
+For binary operators, the type of both operands has to match, expect for exponentiation; an exponent for a value of type `BigInt` always has to be of type `Int`.
```
alan-geller, line 6:
```suggestion
For binary operators, the type of both operands has to match, except for exponentiation; an exponent for a value of type `BigInt` always has to be of type `Int`. 
```

Specifications/Language/3_Expressions/ArithmeticExpressions.md
```diff
@@ -1,7 +1,11 @@
 # Arithmetic Expressions
 
 Arithmetic operators are addition (`+`), subtraction (`-`), multiplication (`*`), division (`/`), negation (`-`), exponentiation (`^`). 
-They can be applied to operands of type `Int`, `BigInt`, or `Double`. Additionally, for integral types (`Int` and `BigInt`) an operator computing the modulus (`%`) is available. For binary operators, the type of both operands has to match, expect for exponentiation; an exponent for a value of type `BigInt` always has to be of type `Int`. The type of the entire expression matches the type of the left operand. 
+They can be applied to operands of type `Int`, `BigInt`, or `Double`. Additionally, for integral types (`Int` and `BigInt`) an operator computing the modulus (`%`) is available. 
+
+For binary operators, the type of both operands has to match, expect for exponentiation; an exponent for a value of type `BigInt` always has to be of type `Int`. 
+The type of the entire expression matches the type of the left operand. 
+For exponentiation of `Int` and `BitInt`, the behavior is undefined if the exponent is negative or if it requires more than 32 bits to represent (i.e. it is larger than 2147483647).
```
alan-geller, line 8: Undefined, or a runtime failure?

alan-geller (approved):

Specifications/Language/3_Expressions/ArithmeticExpressions.md
bettinaheim: Well, in our C# runtime having an exponent with more than 32 bits is actually supported for `Int`. Given that we also leave it at "undefined" for bitshifts by more than 64 bits, I think that is fine here, too.

alan-geller (approved):

PR #35, opened 2020-09-29
QIR: Limit integer exponents to i32

### LINKED ISSUE #28: QIR: reconsider requiring __quantum__rt__int_power(i64, i64) method
Instead of requiring the RT vendors to implement fast full precision integer exponents, consider adding runtime checks to generated QIR that the arguments can be losslessly converted to the types supported by one of the llvm's intrinsic powi/pow functions.

alan-geller:
Closes #28


irinayat-MS (approved):
LGTM. Thanks!

PR #36, opened 2020-09-29
Added message runtime function

### LINKED ISSUE #32: QIR: There is no way to inspect/view/pring %String type
The RT specifies quite a few ways to create an instance of %String type but the only method that might conceivably expose the content of the string to the user is __quantum__rt__fail(%String). Shouldn't there be a standard way for the user to inspect/print a %String?

alan-geller:
Closes #32 

Sorry, this runtime function got lost somewhere along the way.


irinayat-MS (approved):
LGTM. Thanks!

PR #41, opened 2020-10-08
[WIP]: Proposal for allocatable types and generalization of initializers

bettinaheim:
This proposal is work in progress. I am putting up a draft for early feedback and discussion.

### DIFF
--- a/Approved/Initializers.md
+++ b/Approved/Initializers.md
@@ -73,9 +73,6 @@ Initializers built by elevating adjointable operations:
 
 ## Current Status
 
-The right hand side of the assignment in using- and borrowing-statements consists of an initializer or a tuple of initializers. Curre
-
-
 TODO:   
 Describe all aspects of the current version of Q# that will be impacted by the proposed modification.     
 Describe in detail the current behavior or limitations that make the proposed change necessary.      
@@ -83,14 +80,6 @@ Describe how the targeted functionality can be achieved with the current version
 Refer to the examples given below to illustrate your descriptions. 
 
 ### Examples
-
-```qsharp
-initializer BigEndian(nrQs : Int) : BigEndian {
-
-    initialize register = Qubits(nrQs);
-    return BigEndian(register);
-}
-```
 
 Example 1:    
 TODO: insert title and caption
@@ -126,39 +115,7 @@ TODO: insert title and caption
 TODO:   
 Add more examples following the structure above. 
 
-
-Example 2: Nesting of initializer expressions and potential future syntax enhancements   
-Introducing additional syntax sugar for particular operations, such as e.g. a coherent version of a logical `AND` would allow to express oracles constructed from such operators in the following way, if this suggestion is adopted:
-
-```qsharp
-use res = q1 <and> q2 <and> q3; 
-```
-where the right hand side is syntactic sugar ` init And(init And(q1, q2, _), q3, _)`, and q1, q2, q3 are existing qubits that are in scope. If quantum operations are applied to the a qubit that is used as part of the initializer, then an additional qubit is required to preserve the state of that qubit at initialization time. Supposed this is the case for `q1` but not for `q2` or `q3`, then the statement above would be translated into the following upon compilation:
-
-```qsharp
-use res = Qubit();
-use __q1__ = Qubit() {
-    within {
-        CNOT(q1, __q1__);
-        use __var1__ = Qubit() {
-            within { And (__q1__, q2, __var1__); } 
-            apply { And (__var1__, q3, res); }
-        }
-    }
-    apply {
-        // code within the allocation scope
-    }
-}
-```
-
-
 # Implementation
-
-All of the proposed concepts can and should be translated into the existing representation, such that no additional support within the intermediate representation or runtime is needed. The code examples above illustrate how this translation should look like. 
-
-Do we need a special allocation statement for initializers? probably yes; would be nice
-
-All qubits that are in scope and used as part of and initializer need to be "copied out" -> is that possible?? 
 
 TODO:    
 Describe how the made proposal could be implemented and why it should be implemented in this way.    
@@ -166,8 +123,6 @@ Be specific regarding the efficiency, and potential caveats of such an implement
 Based on that description a user should be able to determine when to use or not to use the proposed modification and how.
 
 ## Timeline
-
-The amount of work to implement this is significant. The implementation of the concepts discussed in this proposal will likely happen in stages. 
 
 TODO:    
 List any dependencies that the proposed implementation relies on.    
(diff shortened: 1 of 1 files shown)


Approved/Initializers.md
```diff
@@ -0,0 +1,175 @@
...
+
+### *Allocatable types*:   
+A type is allocatable if it either is one of the built-in allocatable types, or if it contains one or more items which are allocatable. There are two built-in allocatable types: the type Qubit and the type Qubit[], or more generally n-dimensional rectangular Qubit arrays (as suggested [here](https://github.com/microsoft/qsharp-language/issues/39)). 
+
+### *Initializers for allocatable types*:    
+Initializers are special callables that return values of an allocatable type. Initializers are *neither* operations nor functions. They are their own distinct type with no subtyping relation to operation or function types; they cannot be used when a value of operation or function type is required. Initializers can only be used as part of [qubit allocation statements](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/QuantumMemoryManagement.md#quantum-memory-management).
```
cgranade, line 31: Are initializers first-class values (i.e.: can I assign an initializer to a variable binding)?

Approved/Initializers.md
```diff
@@ -0,0 +1,175 @@
...
+An initializer expressions consists either of a call to an initializer, or of a tuple of multiple calls. ...
+
+### Overview over different kinds of initializer expressions
+
+Initializers for built-in types:
+| Built-in Initializer | Description | Allocated Type | Initializer Argument Type | Initializer Expression | 
+| --- | --- | --- | --- | --- |
+| Qubit | allocates a single qubit | `Qubit` | `Unit` | `Qubit()` |
+| Qubits | allocates a 1D array of qubits | `Qubit[]` | `Int` | `Qubits(4)` |
+| Qubits | allocates a [2D array](https://github.com/microsoft/qsharp-language/issues/39) of qubits | `Qubit[,]` | `(Int, Int)` | `Qubits(2,4)` |
```
cgranade, line 63: Effectively, it seems this is a kind of overloading, in that `Qubits` can either have type `Int ~> Qubit[]` or `(Int, Int) ~> Qubit[,]`. Is that represented by a type parameter to `Qubits`?

Approved/Initializers.md
```diff
@@ -0,0 +1,175 @@
...
+### *Initializer expressions and nesting of initializers*:    
+An initializer expressions consists either of a call to an initializer, or of a tuple of multiple calls. ...
+
+### Overview over different kinds of initializer expressions
+
+Initializers for built-in types:
+| Built-in Initializer | Description | Allocated Type | Initializer Argument Type | Initializer Expression | 
+| --- | --- | --- | --- | --- |
+| Qubit | allocates a single qubit | `Qubit` | `Unit` | `Qubit()` |
+| Qubits | allocates a 1D array of qubits | `Qubit[]` | `Int` | `Qubits(4)` |
```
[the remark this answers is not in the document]
cgranade: If I understand correctly, this does not modify array allocation (see https://github.com/microsoft/qsharp-compiler/issues/589 for that proposal), but only the syntax for initializing registers of qubits.

Approved/Initializers.md
```diff
@@ -0,0 +1,175 @@
...
+| --- | --- | --- | --- |
+| BigEndian | `Qubit[]` | `Int` | `BigEndian(4)` |
+| QubitArray | `(Int, Qubit[])` | `(Int, Int)` | `QubitArray(4, 4)` |
+
+Initializers built by elevating operations:
+| Operation Valued Expression | Allocated Type | Initializer Argument Type | Initializer Expression | Uncompute upon Release |
+| --- | --- | --- | --- | --- |
+| `H` | `Qubit` | `Unit` | `init within H` | Yes |
+| `Reset` | `Qubit` | `Int` | `init then Reset` | No |
+| `ApplyToEachA(H, _)` | `Qubit[]` | `Int` | `init(3) within ApplyToEachA(H, _)` | Yes |
```
cgranade, line 76: If it's in scope, it may be helpful to use `ApplyXorInPlace` as an example, since that demonstrates how `init`/`then` can be used to really help out with preparing structured data. In particular, I think that builds well on the example at line 68.

Approved/Initializers.md
bettinaheim: So far I was not planning to support that - they could literally only be used within allocation statements. I don't see a reason to support it at this time. You can always pass an operation that is elevated though. Do you have a use case in mind where you would want to be able to pass an initializer?

cgranade: Makes sense, wasn't trying to suggest that we need first-class initializers per se so much as to understand the scope of the proposal. Thanks for the clarification!

Approved/Initializers.md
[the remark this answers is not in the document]
viktorveis: Sounds good to me. Since we're discussing it here, I would like to better understand advantage of Qubits(4) over Qubit[4].

Approved/Initializers.md
```diff
@@ -0,0 +1,175 @@
...
+
+### *Custom initializers for user defined types:*    
+Support for defining custom initializers for user defined types is intended to be added if and when custom constructors can be defined. Important considerations for what can and cannot be supported for such custom initializers, however, are part of this proposal to ensure that the proposed additions can be extended to cover that scenario as well.    
+In contrast to operations, initializers allocate and return qubits that remain live after their execution terminates. Such qubits are allocated via a dedicated statement, consisting of the keyword `initialize` followed by a symbol or symbol tuple, an equals sign `=`, and either an initializer or an initializer tuple.
```
samarsha, line 40: Having two keywords `init` and `initialize` that mean different things seems confusing to me. It's similar to the `fun` and `function` distinction in F#, which I also think is not very intuitive.

Approved/Initializers.md
```diff
@@ -0,0 +1,175 @@
...
+1. One is that an operation should be invoked on the allocated qubits after initialization, followed by further computations before the qubits are measured and released. In this case, it is not necessary or desirable that the operation performed upon initialization is un-computed upon release. 
+2. In the second case, the computations following the initialization lead to the qubits being in the same state as they were initialized. In this case, the adjoint of the operation used for initialization needs to be applied upon releasing the qubits. 
+
+The proposal is to syntactically distinguish the two cases. The table below contains examples for both cases. In the first case, the initializer consists of the keyword `init`, followed by the initializer argument tuple, the keyword `then` and an operation-valued expression. The initializer argument tuple may be omitted if the required argument is of type `Unit`. The second case follows the same pattern with the exception that instead of the keyword `then`, then keyword `within` should be used, and the operation-valued expression needs to support the `Adjoint` functor.
```
samarsha, line 49: > The initializer argument tuple may be omitted if the required argument is of type `Unit`.

We don't allow omitting `()` when calling functions or operations. What is the reason to support it here?

samarsha, line 49:
```suggestion
The proposal is to syntactically distinguish the two cases. The table below contains examples for both cases. In the first case, the initializer consists of the keyword `init`, followed by the initializer argument tuple, the keyword `then` and an operation-valued expression. The initializer argument tuple may be omitted if the required argument is of type `Unit`. The second case follows the same pattern with the exception that instead of the keyword `then`, the keyword `within` should be used, and the operation-valued expression needs to support the `Adjoint` functor.   
```

Approved/Initializers.md
samarsha, line 31: Having three different types of callable in Q# is a lot, especially since initializer names overlap with both type names and type constructor names. The automatic conversion from e.g. `LittleEndian(Qubit[])` to `LittleEndian(Int)` is also a little confusing - when I see `LittleEndian(3)`, I have to remember that:

1. It's in an initializer expression, so `LittleEndian` refers to the initializer, not the constructor, which takes a different argument type.
2. The `Int` maps back to the length of a `Qubit` array, even though `Qubit` doesn't appear anywhere in the expression.

I think there is a lot of cognitive overhead needed to interpret this syntax.

Approved/Initializers.md
samarsha: The overloading on tuple arity seems weird to me - it's compiler magic that can't be expressed in the language itself. It would not be possible to write a custom initializer that does this. A type parameter would allow any type, not just tuples of `Int`s.

PR #43, opened 2020-10-21
Lots of expansion on dealing with %Callables

alan-geller:
A lot more information on what the runtime actually has to do to handle this structure,

### DIFF
--- a/Specifications/QIR/Callables.md
+++ b/Specifications/QIR/Callables.md
@@ -50,6 +50,18 @@ LLVM type.
 The name of a wrapper function should be the same as the name of the function
 it is wrapping, with "__wrapper" appended.
 
+The four specializations of a callable are:
+
+- The "body", which is the normal, unmodified callable.
+- The "adj", which implements the adjoint of the quantum operation defined by
+  the callable.
+- The "ctl", which implements the controlled version of the quantum operation
+  defined by the callable.
+- The "ctladj", which implements the adjoint of the controlled version.
+
+A callable may define a "ctladj" specialization if and only if it defines both
+"adj" and "ctl" specializations.
+
 There is no need to create wrappers for callables that are never pointed to.
 That is, **a callable that is never turned into a callable value doesn't need wrapper functions**.
 
@@ -183,7 +195,7 @@ The Adjoint and Controlled functors are important for expressing quantum
 algorithms.
 They are implemented by the `__quantum__rt__callable_make_adjoint` and
 `__quantum__rt__callable_make_control` runtime functions, which update a
-`%Callable` in place by applying the Adjoint or Controlled functors
+`%Callable` in place by applying the `Adjoint` or `Controlled` functors
 respectively.
 
 To support cases where the original, unmodified `%Callable` is still needed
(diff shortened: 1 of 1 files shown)


Specifications/QIR/Callables.md
```diff
@@ -205,9 +217,80 @@ The following snippet of LLVM code could be generated:
...
+For instance, if the base callable expects `{ %TupleHeader, %Qubit* }`,
+the singly-`Controlled` version expects `{ %TupleHeader, %Array*, %Qubit* }`
+rather than `{ %TupleHeader, %Array*, { %TupleHeader, %Qubit* }* }`.
+This means that the second element of the singly-`Controlled` argument tuple
+is not always a pointer to a struct, and in particular may have variable length
+up to the size of a `%Range`.
+
+To resolve this, `__quantum__rt__callable_invoke` needs to have access to the
+length of the inner argument tuple once it has unwrapped down to that point.
+This could be stored in the `%TupleHeader` by `__quantum__rt__tuple_create`,
```
irinayat-MS, line 282: Currently, tuple header is prescribed to be `%TupleHeader = type { i32 }`, where the i32 is used to store reference counting.

Specifications/QIR/Callables.md
```diff
@@ -205,9 +217,80 @@ The following snippet of LLVM code could be generated:
...
+Each application of the `Controlled` functor modifies the signature of the
+specialization by adding replacing the current argument tuple with a two-tuple
+containing the array of control qubits as the first element and a tuple of
+the remaining arguments as the second tuple.
+
+For instance, if the base callable expects an argument tuple
+`{ %TupleHeader, i64, %Qubit* }`, then the `Controlled` version expects
+`{ %TupleHeader, %Array*, { %TupleHeader, i64, %Qubit* }* }`, and the twice-`Controlled`
+version expects
+`{ %TupleHeader, %Array*, { %TupleHeader, %Array*, { %TupleHeader, i64, %Qubit* }* }* )`.
```
irinayat-MS, line 262: typo: closing ')' -> '}'

Specifications/QIR/Callables.md
alan-geller: There are 4 different solutions that come to mind (and I'm sure there are others as well):

1. An implementation could use the bottom 24 bits of the `i32` as the reference count and the high 8 bits as the size in bytes; the size is never greater than 24 (for a `%Range`). Admittedly this is kind of yucky.
1. An implementation could use the `%TupleHeader` i32 to store the size of the tuple and have the `tuple_create` runtime function allocate an extra `i32` at the end of the memory block to hold the actual reference count.
1. An implementation could always allocate an extra 'i32' at the start of the memory block when creating a tuple, store the tuple data size there, and return a pointer past that `i32` from `tuple_create`. When the tuple's reference count goes to zero, the pointer to be freed would have to be calculated by adjusting the pointer to the tuple.
1. An implementation could run an LLVM pass that redefines the `%TupleHeader` structure to something such as `{ i16, i8 }` before further compilation.

Do you think it would be helpful to include some or all of these in the specification?

Specifications/QIR/Callables.md
alan-geller: Thanks! Fixed.

Specifications/QIR/Callables.md
irinayat-MS: The solutions are standard enough for most people to be aware of them, my concern was that, as worded, it might confuse readers into thinking that %TupleHeader is extensible. It might help to add into the section on %TupleHeader that the implementers are free to use the i32-sized memory however they please (that is, QIR will never try to interpret this i32). Only 8 bits for reference counting seems too low, though. What are the arguments against extending %TupleHeader to {i32, i32} with explicit expectation that one int is reference count and the other is size, or keeping it unspecified but increasing to i64? Solutions 2&3 would use as much memory anyway.

Specifications/QIR/Callables.md
```diff
@@ -205,9 +217,80 @@ The following snippet of LLVM code could be generated:
...
+The actual implementation of the `%Callable` needs to support the following
+behavior:
+
+- For `__quantum__rt__callable_make_adjoint`, the parity of the number of
+  times this function has been applied should be tracked. The `Adjoint` functor
+  is its own inverse, so applying it twice is the same as not applying it at all.
+  Applying this function to a `%Callable` whose implementation table has a null
+  "adj" entry should cause a runtime failure.
+- For `__quantum__rt__callable_make_control`, the count of the number of 
+  times the function has been applied must be tracked. The `Controlled` functor
```
bettinaheim, line 232:
```suggestion
  times the function has been applied must be tracked, since each application adds an additional item to the argument (explained in more detail below). The `Controlled` functor
```

Specifications/QIR/Callables.md
```diff
@@ -205,9 +217,80 @@ The following snippet of LLVM code could be generated:
...
+One additional complexity is that the above is modified slightly if the base
+callable expects an argument tuple with exactly one element.
+In this case, the `Controlled` version expects a two-element tuples as
+above, but with the actual base argument as the second element.
+For instance, if the base callable expects `{ %TupleHeader, %Qubit* }`,
+the singly-`Controlled` version expects `{ %TupleHeader, %Array*, %Qubit* }`
+rather than `{ %TupleHeader, %Array*, { %TupleHeader, %Qubit* }* }`.
+This means that the second element of the singly-`Controlled` argument tuple
+is not always a pointer to a struct, and in particular may have variable length
+up to the size of a `%Range`.
```
bettinaheim, line 278: Is it clear how things should work if the base argument type is Unit (admittedly a rather horrible edge case that likely is not of real use)?

bettinaheim (approved):
The expansion is a good addition!

Issue #42, opened 2020-10-20
BigInt literals now accept binary and octal formats


samarsha:
microsoft/qsharp-compiler#673 added support for binary and octal BigInt literals to the compiler. The documentation on [BigInt Literals](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/3_Expressions/ValueLiterals.md#bigint-literals) in the Q# specification should be updated.

samarsha:
@ziqi-ma in case you also wanted to update the documentation. :) If not, I or someone else can get to it soon.

ziqi-ma:
Sent in PR:)

PR #44, opened 2020-10-23
update documentation-BigInt now supports binary and octal as well

ziqi-ma:
addressing #42

### DIFF
--- a/Specifications/Language/3_Expressions/ValueLiterals.md
+++ b/Specifications/Language/3_Expressions/ValueLiterals.md
@@ -20,13 +20,13 @@ Value literals for the [`Int` type](https://github.com/microsoft/qsharp-language
 
 ## BigInt Literals
 
-Value literals for the [`BigInt` type](https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/4_TypeSystem#available-types) are always postfixed with `L` and can be expressed in decimal, binary, octal or hexadecimal representation. Literals expressed in binary are prefixed with '0b', in octal prefixed with '0o', in hexadecimal prefixed with `0x`. There is no prefix for the commonly used decimal representation.
+Value literals for the [`BigInt` type](https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/4_TypeSystem#available-types) are always postfixed with `L` and can be expressed in binary, octal, decimal, or hexadecimal representation. Literals expressed in binary are prefixed with `0b`, with `0o` for octal, and with `0x` for hexadecimal. There is no prefix for the commonly used decimal representation.
 
 | Representation | Value Literal |
 | --- | --- | 
+| Binary | `0b101010L` | 
+| Octal | `0o52L` |
 | Decimal | `42L` |
-| Binary | '0x101' |
-| Octal | '0o345' |
 | Hexadecimal | `0x2aL` |
 
 ## Double Literals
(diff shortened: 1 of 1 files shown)


Specifications/Language/3_Expressions/ValueLiterals.md
```diff
@@ -20,11 +20,13 @@ Value literals for the [`Int` type](https://github.com/microsoft/qsharp-language
 
 ## BigInt Literals
 
-Value literals for the [`BigInt` type](https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/4_TypeSystem#available-types) are always postfixed with `L` and can be expressed in decimal or hexadecimal representation. Literals expressed in hexadecimal are prefixed with `0x`. There is no prefix for the commonly used decimal representation.
+Value literals for the [`BigInt` type](https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/4_TypeSystem#available-types) are always postfixed with `L` and can be expressed in decimal, binary, octal or hexadecimal representation. Literals expressed in binary are prefixed with '0b', in octal prefixed with '0o', in hexadecimal prefixed with `0x`. There is no prefix for the commonly used decimal representation.
```
samarsha, line 23: > can be expressed in decimal, binary, octal or hexadecimal representation

Can we use the same ascending order (binary, octal, decimal, hexadecimal) as the section on Int literals above, for consistency?

Specifications/Language/3_Expressions/ValueLiterals.md
```diff
@@ -20,11 +20,13 @@ Value literals for the [`Int` type](https://github.com/microsoft/qsharp-language
...
 
 | Representation | Value Literal |
 | --- | --- | 
 | Decimal | `42L` |
+| Binary | '0x101' |
+| Octal | '0o345' |
```
samarsha, line 29: For consistency with the Int literals section, can we use the same table numbers in the same order? Also, it looks like the binary and octal examples are missing the `L` suffix.

Specifications/Language/3_Expressions/ValueLiterals.md
samarsha, line 23: > Literals expressed in binary are prefixed with '0b', in octal prefixed with '0o', in hexadecimal prefixed with `0x`.

Also related to consistency, the Int literal section above uses the format "with `0o` for octal, and with `0x` for hexadecimal", can we keep that format here as well? By the way, the backticks around `` `0x` `` format it in a monospace font, so I think it'd be good to keep that also, instead of using single quotes around '0b' and '0x'.

samarsha (changes requested):
Thanks for the PR! Since the sections on Int literals and BigInt literals are very similar, I think it would be good to make the BigInt section more consistent with the Int section regarding wording and formatting.

Specifications/Language/3_Expressions/ValueLiterals.md
ziqi-ma: addressed

Specifications/Language/3_Expressions/ValueLiterals.md
ziqi-ma: addressed

Specifications/Language/3_Expressions/ValueLiterals.md
ziqi-ma: addressed

samarsha (approved):
Looks great, thanks!

Issue #46, opened 2020-10-26
Remove "new T[n]" syntax for creating arrays


samarsha:
**Suggestion**

Q# supports two distinct syntaxes for creating arrays: square brackets like `[1, 2, 3]`, and the `new` keyword like `new Int[3]`. When an array is created using the `new` keyword, all elements are initialized to a default value based on the type. The syntax to create arrays using `new` should be removed from the language.

*Considerations*

The existence of `new T[n]` has negative effects on the language.

The most important negative effect is the assumption that every type has a default value. This is not a reasonable assumption, because it is not possible to define useful default values of types like `Qubit` and `T1 -> T2`. Their current default values are invalid and will trigger an error if they are used. Yet the syntax is the same both for types with valid default values, and those without, which is misleading: while `new Int[3]` is a perfectly well-defined value whose elements can be used immediately, `new Qubit[3]` creates a potential trap that will crash the program if any value in the array is used.

This assumption is also invalid in the presence of uninhabited types such as `Void`. If Q# requires that every type has a default value, then uninhabited types are impossible to express properly. This can even cause bugs in the soundness of a type system, as demonstrated by [this bug in Java's generics](https://hackernoon.com/java-is-unsound-28c84cb2b3f) that happened because of the existence of a value (specifically `null`) for a type that should have been uninhabited.

Q#'s default values also caused a subtle bug in the Q# runtime: because the default values required by Q# do not always match the default values of the corresponding C# types (according to C#'s `default` keyword), it is possible to create `null` values of types like `Result`, `String`, and even `Unit`. See microsoft/qsharp-runtime#359.

Finally, the use of the `new` keyword is inconsistent with other ways to create values in Q#. `new` is required only when creating a default-initialized array, not when using a square bracket array literal or when creating values of other types.

*Context*

See microsoft/qsharp-compiler#589 for a more detailed discussion about default values and alternatives to the `new T[n]` syntax. 

*Examples*

There are functions in the standard library that can create arrays instead of using the `new T[n]` syntax.

```qsharp
// Instead of:
let empty = new Int[0];
let results = new Result[5];

// Use:
let empty = EmptyArray<Int>();
let results = ConstantArray(5, Zero);
```

**Affidavit (please fill out)**

Please add ticks by placing a cross in the box:
* [x] I have searched both open and closed suggestions and proposals on this site and believe this is not a duplicate.
* [x] I believe that the spirit of this suggestion is aligned with the design principles and general vision for Q#. 

[...]

PR #48, opened 2020-10-28
Proposal for removing default-initialized new array expressions

samarsha:
This is the proposal for issue #46.


Approved/remove-new-array.md
```diff
@@ -0,0 +1,84 @@
...
+
+```qsharp
+// Create an empty arra of type Int[].
+let empty = EmptyArray<Int>();
+
+// Create an array of length 10 of type Bool[] where every value is false.
+let bools = ConstantArray(10, false);
+```
+
+# Implementation
```
bettinaheim, line 55: I believe EmptyArray and ConstantArray right now are not intrinsic functions. We would need either a special syntax or an intrinsic function that connects to the corresponding runtime functionality.

Approved/remove-new-array.md
```diff
@@ -0,0 +1,84 @@
...
+
+N/A
+
+## Anticipated Interactions with Future Modifications
+
+Improved type inference will allow the empty array to be expressed as `[]` instead of `EmptyArray<a>()`.
+
+## Alternatives
+
+N/A
```
bettinaheim, line 80: I think it is worthwhile to consider whether it makes sense to introduce syntax sugar for common array instantiations, such as ConstantArray. I think this might make sense here.

Approved/remove-new-array.md
cgranade: Quick side note, but while ConstantArray is implemented in Q#, EmptyArray is already intrinsic in the runtime (https://github.com/microsoft/qsharp-runtime/blob/a18ecf4eb1a6cad4800c6dfe4a5ce96e362d807c/src/Simulation/QsharpCore/Arrays/Empty.cs).

Approved/remove-new-array.md
samarsha: Do you have a concrete example of syntax for a constant array? I can't think of anything reasonable.

Approved/remove-new-array.md
```diff
@@ -0,0 +1,84 @@
...
+// Create an empty arra of type Int[].
+let empty = EmptyArray<Int>();
+
+// Create an array of length 10 of type Bool[] where every value is false.
+let bools = ConstantArray(10, false);
+```
+
+# Implementation
+
+The library functions `EmptyArray` and `ConstantArray` are already implemented.
```
viktorveis, line 57: Does this mean that we'll need to import a library every time we create an array?

Approved/remove-new-array.md
```diff
@@ -0,0 +1,84 @@
...
+The syntax above will be removed.
+The same functionality can be accomplished with functions in the standard library:
+
+```qsharp
+// Create an empty arra of type Int[].
+let empty = EmptyArray<Int>();
+
+// Create an array of length 10 of type Bool[] where every value is false.
+let bools = ConstantArray(10, false);
+```
```
viktorveis, line 53: I agree that removing the `new` keyword is goodness. However, the proposed syntax may feel foreign for developers coming from Python, C#, Java, or C++.  It will also look inconsistent with `using (a = Qubit[10]) {}`. We should explore syntax to create empty and constant arrays that is consistent with both `a = [1, 2, 3];` and the qubit example.  For example, Python has an elegant syntax for these scenarios:
```
a = []
b = [1] * 10
```

Approved/remove-new-array.md
samarsha: You can still create array literals like `[1, 2, 3]` without library functions. To create empty arrays or arrays consisting only of a repeated constant, you need to use `EmptyArray` or `ConstantArray` from the `Microsoft.Quantum.Arrays` namespace. We could consider moving them into `Microsoft.Quantum.Core` so that it's not necessary to `open Microsoft.Quantum.Arrays` to access them.

Approved/remove-new-array.md
samarsha: > However, the proposed syntax may feel foreign for developers coming from Python, C#, Java, or C++.

To clarify, this proposal does not contain any new syntax. `EmptyArray` and `ConstantArray` are functions that are already defined in the Q# standard library. They have counterparts in C# too:

```csharp
var a = Array.Empty<int>();
var b = Enumerable.Repeat(1, 10).ToArray();
```

> ```
> a = []
> ```

Yes, this is my favorite way to create an empty array, but it's blocked for the moment because Q#'s type inference is not sufficient to support this. See microsoft/qsharp-compiler#617.

> ```
> b = [1] * 10
> ```

Because Q# is statically typed, unlike Python, this syntax has some additional implications for us:

* The type of `*` is overloaded to both `('a[], Int) -> 'a[]` and `(Int, 'a[]) -> 'a[]`, since Python also supports `10 * [1]`. Q# does not support overloading. I would argue that there are good reasons for Q# to never support overloading, but there is a feature request for this: microsoft/qsharp-compiler#29
* We would need to choose a different operator name than `*`. Otherwise, it's not possible to reconcile the type of `*` used for arrays with the type of `*` needed to multiply two numbers together. This would make it very difficult to create a type class to represent numeric types. See microsoft/qsharp-compiler#557.

Because of these issues, and because in my opinion, `[1] * 10` is harder to read than `ConstantArray(10, 1)`, I would recommend keeping `ConstantArray` as a normal function instead of adding new syntax to Q#.

viktorveis: Having these library functions is fine. I'm concerned with making them the only way to create certain kind of arrays. I don't see how Q# developers will be excited to go through their code to replace `a=new Int[0]` with `a = EmptyArray<Int>()`. It is verbose and not common in other languages. In contrast, replacing `a=new Int[0]` with `a = []` feels like a great change. I think that we should hold on deprecating the `new` keyword until microsoft/qsharp-compiler#617 is fixed. Requiring to replace `a = new Int[1]` with `a = ConstantArray(1, 0)` isn't great either.  We don't have to use the python syntax as is but rather explore similar ideas as alternative options to ConstantArray.

samarsha: > I think that we should hold on deprecating the `new` keyword until microsoft/qsharp-compiler#617 is fixed.

While I agree that `[]` is the best syntax, and that we should try to avoid deprecating syntax in favor of alternatives that will themselves be replaced, we also need to balance the benefits of removing default value semantics from Q# with the effort needed to fix microsoft/qsharp-compiler#617.

1. I think it's important to remove default values by Q# 1.0. If they exist in 1.0, they'll be harder to remove later. By contrast, it's much easier to add `[]` and/or a built-in constant array syntax later on.
2. The effort needed to fully resolve microsoft/qsharp-compiler#617 is large. We could fix some limited cases more easily, but we'd still need some way to explicitly specify the type of an empty array for the remaining cases. There's alternative syntax we could consider instead of `EmptyArray<T>()`, such as `[] : T[]`.
3. The alternative that we pick for `new T[0]` should persist even after `[]` is added - I don't expect `EmptyArray` or `ConstantArray` to ever be deprecated, so while we'll be providing a better alternative eventually, there won't be another round of deprecation like there was with `new`.

> It is verbose and not common in other languages.

For comparison, `Array.Empty<T>()` is preferred over `new T[0]` to create empty arrays in C#: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1825 (albeit for different reasons than Q#).

> Requiring to replace `a = new Int[1]` with `a = ConstantArray(1, 0)` isn't great either. We don't have to use the python syntax as is but rather explore similar ideas as alternative options to ConstantArray.

I would be interested in concrete syntax suggestions. I cannot think of any, and other than Python, I am not aware of any language that provides this functionality using operator syntax instead of a named function. I also think this is subjective, and not a clear-cut case where special syntax is definitely needed. The function seems preferable to me, other than this specific case of an `Int` array where it's not immediately clear which parameter is the length and which is the value.

Approved/remove-new-array.md
samarsha: @cgranade, do you have any thoughts on adding `EmptyArray` and/or `ConstantArray` to `Microsoft.Quantum.Core`?

Approved/remove-new-array.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Remove default-initialized new array expressions
+description: The `new Type[N]` array expression is deprecated and removed from Q#.
+author: Sarah Marshall
+date: October 28, 2020
+---
+
+# Proposal
+
+The `new Type[N]` expression, which creates an array of `N` values using the default value of `Type`, is deprecated and removed from Q#.
```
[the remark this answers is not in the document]
samarsha: Added with slightly different wording.

Approved/remove-new-array.md
samarsha: Added some more about this. Technically, `ConstantArray` doesn't need to be intrinsic, but it could be more efficient if it was, so I added a note that we should do that.

Approved/remove-new-array.md
cgranade: To clarify, with the current state of affairs, no library needs to be imported to use `EmptyArray` and `ConstantArray`, since both are part of the Q# standard library and hence provided by the packages made available  by default in new projects (whether Q# standalone command-line, Q# standalone notebooks, Q# + Python, or Q# + .NET). That said, one does need an `open` statement at the moment to make the `Microsoft.Quantum.Arrays` _namespace_ available; our language server already suggests doing so if a user has forgotten in VS Code or VS 2019.

With respect to moving `EmptyArray` and `ConstantArray` to the Microsoft.Quantum.Core namespace, I'm somewhat leery since everything in Microsoft.Quantum.Core is automatically opened and thus available to all Q# programs even without an `open` statement. That's a significant exception to the use of namespaces to document where things come from, organize code, and so forth — that exception is really helpful in cases like `Length`, but in general I think we should make as few exceptions of that form as is reasonable. Thus, while it may make sense to move those two functions, I'd suggest that before doing so, we expand the API design principles to be more clear about the role of the Microsoft.Quantum.Core namespace in particular.

samarsha: My ideal solution, which doesn't fit Q#'s naming convention right now, is to auto-open `Microsoft.Quantum.Arrays` using an alias - so every Q# namespace has an implicit

```qsharp
open Microsoft.Quantum.Arrays as Array;
```

Then I would rename `EmptyArray` to `Empty`, and `ConstantArray` to `Constant`, and we could do `Array.Empty<T>()` and `Array.Constant(n, value)`.

cgranade: That feels eerily close to modules...

Approved/remove-new-array.md
cgranade: Irrespective of how this proposal goes more generally, I agree with moving `ConstantArray` to being intrinsic as a performance consideration. If possible, it may be good to keep the pure-Q# fallback to make it easier for custom simulators by reducing how much they have to implement.

Approved/remove-new-array.md
cgranade: > > I think that we should hold on deprecating the `new` keyword until [microsoft/qsharp-compiler#617](https://github.com/microsoft/qsharp-compiler/issues/617) is fixed.
> 
> While I agree that `[]` is the best syntax, and that we should try to avoid deprecating syntax in favor of alternatives that will themselves be replaced, we also need to balance the benefits of removing default value semantics from Q# with the effort needed to fix [microsoft/qsharp-compiler#617](https://github.com/microsoft/qsharp-compiler/issues/617).
> 
> 1. I think it's important to remove default values by Q# 1.0. If they exist in 1.0, they'll be harder to remove later. By contrast, it's much easier to add `[]` and/or a built-in constant array syntax later on.

Absolutely agreed, default values have been a pretty significant thorn in the side of Q#'s type safety for a while now. The sheer breakage one can cause with `H((new Qubit[1])[0]);` is pretty impressive, for instance.

> 2. The effort needed to fully resolve [microsoft/qsharp-compiler#617](https://github.com/microsoft/qsharp-compiler/issues/617) is large. We could fix some limited cases more easily, but we'd still need some way to explicitly specify the type of an empty array for the remaining cases. There's alternative syntax we could consider instead of `EmptyArray<Int>()`, such as `[] : Int[]`.
> 3. The alternative that we pick for `new T[0]` should persist even after `[]` is added - I don't expect `EmptyArray` or `ConstantArray` to ever be deprecated, so while we'll be providing a better alternative eventually, there won't be another round of deprecation like there was with `new`.
> 
> > It is verbose and not common in other languages.
> 
> For comparison, `Array.Empty<T>()` is preferred over `new T[0]` to create empty arrays in C#: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1825 (albeit for different reasons than Q#).
> 
> > Requiring to replace `a = new Int[1]` with `a = ConstantArray(1, 0)` isn't great either. We don't have to use the python syntax as is but rather explore similar ideas as alternative options to ConstantArray.

By way of understanding, what isn't great about it? The `new`-based syntax doesn't make clear what value elements are populated with; the user has to read the table of default values for each type in the Q# language docs to know that `new Int[1]` and `[0]` are equivalent. By contrast, `ConstantArray(3, PauliX)` clearly documents to the _reader_ that the length of the new array is 3, and that each element is `PauliX`. In general, that goes well with the focus that Q# should be easy to read, even when that makes _writing_ Q# slightly more cumbersome; especially given difficulty posed by the problem domain of quantum computing, it's important to make sure that code is as readable and understandable as it can reasonably be.

> I would be interested in concrete syntax suggestions. I cannot think of any, and other than Python, I am not aware of any language that provides this functionality using operator syntax instead of a named function. I also think this is subjective, and not a clear-cut case where special syntax is definitely needed.

As a quick clarification, `[value] * length` in Python is an effect of `+` being used for list concatenation. That is, if one follows common intuitions about algebraic expressions, I don't think that `[a, b] + [a, b]` being the same thing as `[a, b] * 2` is wholly unreasonable  —  indeed, the syntax reflects that Python lists are monoids under `+`. That is, Python doesn't have special syntax for "create a list with 𝑛 copies of the same value" so much as it has a general syntax for "repeat this list 𝑛 times." From that perspective, if we want to borrow Python's `[a] * n` syntax, I would strongly suggest we not treat it as a special case, but that we also treat the monoid concept more generally.

> The function seems preferable to me, other than this specific case of an `Int` array where it's not immediately clear which parameter is the length and which is the value.

We could possibly address that by a named input feature, something like `ConstantArray(length = 10, value = 3)`. Happy to think a bit on that and draft a suggestion if that would be helpful.

Approved/remove-new-array.md
samarsha: Yes, this is one of several reasons why I think Q# should replace namespaces with modules. Having access to all functions related to arrays, lists, etc. in F#, without polluting the global namespace, is something I miss in Q#.

Approved/remove-new-array.md
samarsha: > From that perspective, if we want to borrow Python's [a] * n syntax, I would strongly suggest we not treat it as a special case, but that we also treat the monoid concept more generally.

I'm fine with Python supporting repeating a list multiple times (i.e. repeated application of the monoid operation on lists), my issue is with overloading `*`. :) In general, multiplication is not repeated addition. More concretely it conflicts with the usual type of `* : a -> a -> a`.

cgranade: > > From that perspective, if we want to borrow Python's [a] * n syntax, I would strongly suggest we not treat it as a special case, but that we also treat the monoid concept more generally.
> 
> I'm fine with Python supporting repeating a list multiple times (i.e. repeated application of the monoid operation on lists), my issue is with overloading `*`. :) In general, multiplication is not repeated addition. More concretely it conflicts with the usual type of `(*) : a -> a -> a`.

Fair enough, though that's an abuse perpetuated by group theorists for many years; using `*` to denote repetitions of a monoid addition is an abuse that is not unique to Python. I agree, though, the historical existence of an abuse of notation isn't a great argument for perpetuating it, heh.

Approved/remove-new-array.md
samarsha: I forgot that you can still have a Q# fallback for a callable that has a native implementation. I clarified this in the proposal.

bettinaheim: I don't think having it be intrinsic necessarily impose an additional burden on the runtime beyond what already needs to be supported; the ConstantArray can be recognized and replaced with whatever construct the runtime supports prior to linking. 
@alan-geller for input on whether we want to require such a function being supported in QIR runtime regardless. 
@swernli for guidance on how it best fits with the recent C# runtime restructuring (I believe the way it aligns with what else we have would be to have it be an intrinsic function in the Q# foundation).

Approved/remove-new-array.md
bettinaheim: I agree we don't need to rush it. It likely makes more sense to first introduce multi-dimensional arrays. Then we can consider if and what array manipulation capabilities we would like to support in general, and what syntax sugar might make sense. I think what I was thinking was somewhat aligned with having an operator that given an array, constructs a stack or sequence of x such arrays, but I agree that doesn't necessarily seem that useful more generally. It would still be good to add a sentence here that adding dedicated syntax is something one could consider at this point, but we have decided that it makes more sense to potentially support it later only if there is a more general need for array manipulation operators.

Approved/remove-new-array.md
bettinaheim: I think having to construct a constant array is common enough that the namespace would have to be opened almost everywhere and we should find a better solution. Related, also IndexRange is one thing that we might consider having available without having the explicitly open a namespace. We could certainly mitigate by having the M.Q.Array namespace opened in our project templates like M.Q.Canon. But even then, having an increasingly long list of namespaces at the top of each file even for a small example to me is detrimental to the impression that everything is simple. I hence think we should have more auto-opening than we currently do. 
I am curious about the arguments against having namespaces automatically open under an alias by default, with the option to disable that with a project configuration. @cgranade @guenp @msoeken, thoughts?

Approved/remove-new-array.md
cgranade: Link to multidimensional arrays proposal: https://github.com/microsoft/qsharp-language/pull/49

Approved/remove-new-array.md
```diff
@@ -0,0 +1,92 @@
...
+A native implementation of `ConstantArray` should be added that allocates the whole array before initializing the items with the given value.
+The pure Q# implementation should remain as a fallback in case the native implementation is not available.
+
+## Timeline
+
+N/A
+
+# Further Considerations
+
+## Related Mechanisms
```
bettinaheim, line 74: We should list the current array initializer syntax (`using (a = Qubit[10]) { ... }`) as related, and link the initializer proposal. I think it would indeed be good to get rid of that oddity altogether and make all initializers be call-like.

Approved/remove-new-array.md
bettinaheim: Sorry for coming late to this thread; good discussions!
"I would argue that there are good reasons for Q# to never support overloading" we already do support some overloading, like + that can be used for all arithmetic types and also for concatenations. 
"We would need to choose a different operator name than *." Yes, I agree that this would indeed be good in this case.

I think we all want to support empty array literals (`[]`), and the only question is around priority here. The priority conversation is a separate one to have outside this proposal. What we do need to decide for this proposal here is whether or not having support for empty array literals is a blocker for removing new-array expressions. I don't think it is, but I am happy to chat about whether it should be. 
I wouldn't introduce something like `[] : Int[]` as a "one-off" thing for new arrays only. If we were to introduce something like that syntax, then is should probably be supported for all expressions in general. An alternative to that would be to allow explicit type annotations for local bindings (`let a : Int[] = [];`). Both of these require careful consideration and are more work to implement than supporting `[]` without any annotation. 

"if we want to borrow Python's [a] * n syntax, I would strongly suggest we not treat it as a special case, but that we also treat the monoid concept more generally." Yes, fully agreed.

samarsha: > We already do support some overloading, like + that can be used for all arithmetic types and also for concatenations.

The operators in Q# right now sort of exist outside of the type system, so I don't think it's overloading exactly. The issue I am thinking of is if/when we want to define `*` as a function in Q#, we need to give it a proper type. That's not possible without overloading in the cases of numeric multiplication and list multiplication.

For `+` specifically, depending on how we define it, it doesn't necessarily require overloading. It at least has the same type in both cases: lists are `('a[], 'a[]) -> 'a[]`, which is still an instance of `('a, 'a) -> 'a` that you would expect for numeric `+`. But we may want to make it part of a different type class (Haskell separates them), in which case we would need to either give it a separate name or introduce overloading into the type system proper.

samarsha: > I wouldn't introduce something like `[] : Int[]` as a "one-off" thing for new arrays only.

I agree, and I wouldn't propose that we do that. If we used `[] : Int[]`, I would agree that it should be type annotations for all expressions.

Approved/remove-new-array.md
cgranade: Thanks for pinging me, @bettinaheim! Following our offline discussion, if I could try to summarize, it would be good to have a way to declare aliases as being automatically opened with new projects and notebooks before moving forward with verbose names like `Microsoft.Quantum.Arrays.EmptyArray`. For example, if `Array.Empty<Int>()` worked by default using auto-open functionality, and could be reduced further to `Empty<Int>()` with `open` statements, then that would make the library-based approach a better option. 

Two options for doing so that may be good to include in a new proposal: 

- Allow "re-exporting" namespace aliases declared with `open ... as` statements.
- Allow defining namespace aliases in csproj file properties.

samarsha: > and could be reduced further to `Empty<Int>()` with open statements, then that would make the library-based approach a better option.

In my opinion, `Empty<T>()` is confusing and should be avoided. If we shorten the names to make the aliases read better, I'd like to see Q# adopt something like F#'s `[<RequireQualifiedAccess>]` instead.

Approved/remove-new-array.md
bettinaheim: Let me summarize some of our conversation today. 

Looking in more detail at what operators make sense for matrix multiplication, we think that there is not much need beyond matrix multiplication, broadcasting, and maybe an operator for a tensor product (that can be covered by the broadcasting, but is pretty unreadable that way). Python additionally supports concatenation for multidimensional arrays, but is is unclear whether there would be much value for us to do that same. Out of these, there is no syntax that naturally covers the scenario of array construction. 

We have had an extensive conversation about the value of modules. The clear upside to using modules over free floating callables, leading to an Array.Constant syntax over ConstantArray, is the additional discoverability of callables that can be easily identified via an appropriate module. For callables that should exist not just for arrays but more broadly for collections/enumerables in the future, such as e.g. Mapped, it makes more sense to have them not be part of an Array module since it would imply that we would then in the future have not just Array.Mapped but also List.Mapped, instead of simply a parameterized version of Mapped. In the case of Constant, the argument alone does not make it clear whether an array or e.g. a list should be constructed. However, with the same effort that is needed to support empty array literals `[]`, it is possible to infer the type argument for the return type such that the same considerations as for Mapped are also applicable to Constant. 

Replacing `new Int[10]` simply with `Constant(10)` or e.g. `Repeat(10)` could be an option but given that there is no clear preferred naming that is consistently used across languages we would basically force people to memorize the name, which seems rather suboptimal. 

We did not come to a clear conclusion and path forward despite our best efforts. 
One path we haven't explored that might give some inspiration is to think about whether some limited form of a "yielding loop" expression (similar to `[for i in 1..10 do yield i]` in F#) could be interesting and sufficiently compact without being overly cryptic.

bettinaheim: After thinking some more: 
While I think having something similar like Python's list comprehension might be valuable to avoid the awkwardness if the same thing needs to be expressed with the current library support, any reasonable way of expressing that I can come with is not compact/convenient enough to be a replacement for the new array expressions. However, I think it might make sense to fully buy into and expand on the notion that "things within rectangular brackets are special", and allow for a dedicated syntax to instantiate arrays of a certain length populated with a given value. I am thinking that something like 
`["", length=10]` might be a good option.

samarsha: Honest question, but do we want custom syntax because we think it's more intuitive/easier to read/more discoverable than a function? Or do we want it because we can't come up with a naming convention for the function? :)

IMO, `Array.Constant` is a lot more intuitive and discoverable than any custom syntax could be, because it's plausibly guessable - start typing `Array.` and get completions. And, IMO, it's also more readable - no need to interpret custom syntax invented for a single scenario, but it's simply a normal function call that is used throughout the language.

bettinaheim: Following the summary I gave above my suggestion, having Array.Repeated and then likely at some point in the future e.g. List.Repeated is suboptimal because we are now forced to implement a separate Repeated function for each type rather than having the option to express it in terms of supporting certain kinds of types. It is also awkward to extend to the multidimensional case and requires that we default import Array. `["", size=10]` on the other hand is easy enough to read, extensible to the multidimensional case, and more or less aligned with other syntax sugar we should consider adding for building arrays and similar types in the future.

samarsha: I am not sure why this applies to `Array.Repeated`, but not the rest of the `Microsoft.Quantum.Arrays` namespace. That entire namespace may be deprecated and generalized if/when Q# supports type classes. Why does `ConstantArray` need its own syntax, but not any of the rest of the functions? (Besides `EmptyArray`, but syntax for that naturally falls out of the general array literal syntax.)

bettinaheim: I am not sure I follow? My logic is that indeed that a good part of the M.Q.Arrays namespace won't be specific to arrays and in time hence migrate to a M.Q.Collection or M.Q.Enumerable or similar namespace. Having an Collection.Repeated however doesn't make sense, since it is not clear what Repeated needs to construct given only its argument. The only option hence is to either have an awkward split where functions like `Mapped` etc are in the collection namespace whereas `Repeated` is duplicated in the namespace/module for each type, or have all functions separate for each type (Array.Mapped, List.Mapped, etc), which adds unnecessary verbosity since in these cases the type can be determined by the argument.

samarsha: > Having an Collection.Repeated however doesn't make sense, since it is not clear what Repeated needs to construct given only its argument.

I don't understand this. We are already introducing return type inference to support `[]`. That would mean `Collection.Repeated` could infer the type based on the context in which the return value is used.

samarsha: Also, there are more collection types than we could reasonably invent syntax for. While arrays have literals, lists may or may not have literals in the future, and there are also lazy sequences, dictionaries, etc. Those types will probably need there own `Seq.Repeated`, `Seq.Empty`, `Dict.Empty`, etc. I think it's reasonable to have a type-specific namespace/module for things that weren't generalized to a type class.

bettinaheim: I believe the syntax for the relevant types would be possible to keep pretty much the same. Right on the inferring the type argument based on the usage for Repeat. That would then lead to using Repeat on its own, which brings us back to "it's something people will just need to learn by heart" without a good option to make it easily discoverable.

samarsha: > I believe the syntax for the relevant types would be possible to keep pretty much the same.

For all possible collection types? Even with type classes, it's not feasible to generalize everything. Type-specific modules is a common pattern even in languages that support type classes. I think it's a naming issue we'll need to confront eventually (`open ... as ...` is only sort of an ad-hoc way to do this).

> That would then lead to using Repeat on its own, which brings us back to "it's something people will just need to learn by heart" without a good option to make it easily discoverable.

If `Repeated` is in `Microsoft.Quantum.Core` or another auto-opened namespace, then I think it's still more discoverable than custom syntax due to code completion.

bettinaheim: Having specific namespaces for everything that is specific to a certain type makes perfect sense - I am not arguing there. What I think we should avoid is duplicating things that can be expressed more generally in the namespace for each type. 
I believe we both agree that we are down to the option of either having array creation be expressed simply as Repeat(10, "") with Repeat being in an automatically opened namespace or having it be custom syntax. Agreed?

samarsha: I'm not sure that `Array.Repeat`, `List.Repeat`, `Seq.Repeat`, etc. should be ruled out yet. I would need some more details about how exactly it should be generalized into a type class. Assuming we have some collection-ish type class, there are two options:

1. `Repeat` is not part of the type class definition. It can be implemented in terms of other collection functions.
2. `Repeat` is part of the type class definition. Every collection must provide an implementation of it.

I could be wrong, but I don't think that 1 is possible. That leaves 2, which means every type that has an instance of the collection type class must implement `Repeat` separately. This doesn't necessarily make sense, depending on how we define our type classes.

For example, it could be reasonable to say that a dictionary is a type of collection. But what does it mean to create a dictionary by repetition? The keys can't be repeated, but if a value is repeated, what key is assigned to each instance of the repeated value?

There may be other ways to organize the type classes that avoids this problem. But I'm not sure it's a given that `Repeat` can or should be generalized.

samarsha: Sets are another example. In the case of sets, it really doesn't make sense to construct a set by repetition. But they can be treated as a collection in most other ways (mapping, folding, etc.).

bettinaheim: Let's just look at what types Repeat should exist for (assuming the existence of all the things that we don't yet have). Repeat should exist for array, list, and sequence. As you point out, there is no meaningful way to define something like "initializer this collection with n times the same value" for sets and dictionaries, and similarly they also don't support e.g. numbers item access. I am not sure why considering set and dictionary to be of the same "kind" of type should be a given? 
You may agree that there is more than one type for which repeat should exist (at least for every type of arbitrary length that supports numbered item access I believe repeat makes sense). There should be a way to express one function called Repeat that can be used with all of those types. If that is not possible to express with type classes, then type classes don't fit the need we have.

samarsha: > There should be a way to express a function called Repeat that can be used with all of those types opposed to having to distinguish the function for each type by name. If that is not possible to express with type classes, then type classes don't fit the need we have.

Type classes are capable of doing this. At worst, you could define a RepeatableCollection for collection types where a `Repeat` could make sense...

But why is this a need? I think this assumption should be examined. It is not necessarily the case that everything in common needs to be generalized. What is the concrete need to write functions that can construct a collection by repetition, that are also polymorphic over the specific collection type?

bettinaheim: Indeed. Suppose that we want to define Repeat in the array namespace. This brings us back to that we have two options: Either make a fairly arbitrary split regarding which functions are in the type specific namespace and which ones are in a more general namespace, or put all functions which are similar to Repeat (e.g. Mapped) into the type specific namespace. 
You are advocating for the latter I believe, and I have some concerns based on my F# usage that does just that. I expect functions such as Mapped to be used fairly often. If they are defined for each type individually, then changing the type e.g. from list to sequence requires updating the code base. In addition to that hassle, having to qualify the name makes it more verbose (writing Array.Mapped instead of merely Mapped).

samarsha: >  Either make a fairly arbitrary split regarding which functions are in the type specific namespace and which ones are in a more general namespace, or put all functions which are similar to Repeat (e.g. Mapped) into the type specific namespace.
You are advocating for the latter I believe

No, I'm not suggesting we duplicate `Array.Mapped`, `List.Mapped`, etc. like F# does. I think `Mapped` should be generalized into a type class.

I'm saying that I don't think having a split between type class functionality and type-specific functionality is something that we should avoid at all costs. In most cases, I think it's OK to require specifying the exact type you want to construct - we already have to do this with array literals like `[1, 2, 3]`, and I don't see a push to generalize that into `MakeIndexedCollection(1, 2, 3)` that can be return-type-inferred into an array. And the syntax you're proposing, `["", size=10]`, already specifies the type to be an array, so it's no worse than `Array.Repeated` in that regard.

samarsha: The syntax you proposed implies some additional possibilities. Can you clarify if you want to support any of them?

* `let n = 4; ["", size = n]`
* `[size = 4, ""]`
* `["foo", "bar", size = 4]` -> `["foo", "bar", "foo", "bar"]` or error?
* `["foo", "bar", size = 3]` -> `["foo", "bar", "foo"]` or error?

Also, I assume you picked `size` instead of `length` to generalize to multidimensional arrays, but this unfortunately means that `Length(["", size = 4]) == 4`, but not `Size(["", size = 4]) == 4` unless we also add a `Size` function.

bettinaheim: Yes, I think it is a fair point to basically associate initialization functions with the type, and all transformation functions with the class of types, and I agree that the syntax solution also associates it with a type. 
I would only support the first one, not anything else.

samarsha: > associate initialization functions with the type, and all transformation functions with the class of types

Unfortunately, I don't think we can cleanly divide it like this, either - see for example `mempty` and `return`, both of which are constructors that are part of a type class:

```haskell
mempty :: Monoid a => a
return :: Monad m => a -> m a
```

My point is mostly that I don't think we need to have a 100% consistent rule here. I think when designing type classes in the standard library, we'll need to make decisions about what's in a type class and what's not, and not everything that can be generalized to more than one type should automatically go into a type class, necessarily. Even if `Repeated` is put into a type class, it's possible that both `SomeCollectionClass.Repeated` and `Array.Repeated` could work, where `Array.Repeated` is just the specific implementation for arrays.

cgranade: To summarize a bit of discussion that @bettinaheim and I had during our sync, I think we were able to come up with a few more ideas that may play nicely with future syntax (e.g.: array comprehensions and multidimensional arrays):

- `["" rep 10]`
- `["" dim 10]`
- `[10 of ""]`
- `["" | size=10]`

I'm personally a bit partial to `[10 of ""]` since it reads closer to English prose ("I want 10 of that"), but all of these generalize nicely to the 𝑛-d case (e.g.: `["" dim (3, 4)] == #[["", "", "", ""], ["", "", "", ""], ["", "", "", ""]]`), and can be made to work nicely with array comprehension. In particular, if we go with the size / shape on the left, then we could have array comprehensions like `[for i in 1..10 -> i * i]` (or `[for i in 1..10 => DrawRandomDouble(0. PI())]` for operations), while if we go with the element being repeated on the left, we could have comprehensions like `[i * i for i in 0..10]` (Python- or Julia-style).

bettinaheim: My slight aversion to `[10 of ""]` (which may just be a personal thing that most people don't perceive that way) is that when I am reading it, it I stumble upon it because starting with `[10` I am already thinking it is an array of integers with an element 10, and only when continuing I need to mentally revise to "no, it is actually an array of strings". That may be silly especially for a tiny piece of code, but suppose both expressions are larger and the expression is spread across multiple lines, it would require a double take for me to mentally parse it. I'm sure I wouldn't even notice that double take after a reading a few of these constructs, but I am wondering if it may be relevant in terms of how natural/easy it feels to use it.

cgranade: > That may be silly especially for a tiny piece of code, but suppose both expressions are larger and the expression is spread across multiple lines, it would require a double take for me to mentally parse it. I'm sure I wouldn't even notice that double take after a reading a few of these constructs, but I am wondering if it may be relevant in terms of how natural/easy it feels to use it.

I don't think it's silly at all; I'll admit I have a slight preference towards `of` since I read more it as prose, but I definitely take the point that it could be hard to read as it gets more complicated. Something like `[(nPlanes, nRows, nCols) of Complex(0.0, 0.0)]` to initialize a three-dimensional array of complex numbers could be a bit hard to read.

viktorveis: I think that `let a = [1 dim (3, 4)];` is compelling. Will this be supported: `let x = 1; let d = 10; let a = [x dim d];`? How will it look like to initialize an array of qubits?

bettinaheim: Yes, that will be supported. 
If `q` is an initialized qubit then `[q dim 5]` would create an array with each item set to that qubit (though I would doubt that such an array is very useful). To get an initialized qubit, you have to use an allocation such as `use qs = Qubit[5];` (though we might want to revise that syntax slightly), just like it is the case currently.

crazy4pi314: > To summarize a bit of discussion that @bettinaheim and I had during our sync, I think we were able to come up with a few more ideas that may play nicely with future syntax (e.g.: array comprehensions and multidimensional arrays):
> 
> * `["" rep 10]`
> * `["" dim 10]`
> * `[10 of ""]`
> * `["" | size=10]`
> 
> I'm personally a bit partial to `[10 of ""]` since it reads closer to English prose ("I want 10 of that"), but all of these generalize nicely to the 𝑛-d case (e.g.: `["" dim (3, 4)] == #[["", "", "", ""], ["", "", "", ""], ["", "", "", ""]]`), and can be made to work nicely with array comprehension. In particular, if we go with the size / shape on the left, then we could have array comprehensions like `[for i in 1..10 -> i * i]` (or `[for i in 1..10 => DrawRandomDouble(0. PI())]` for operations), while if we go with the element being repeated on the left, we could have comprehensions like `[i * i for i in 0..10]` (Python- or Julia-style).

I think from this list either  `["" | size=10]` or `[10 of ""]` are the most clear to me. 
What about `["" size 10]`? Then you get the clear keyword `size` and minimal additional syntax?

bettinaheim: One additional suggestion buried in the thread is also `["", size=10]`. Any thoughts on using `size` also in the multidimensional case? I would want to work is also for that.

samarsha: I like that `size` is a word (`rep` and `dim` are very cryptic to me), but I don't like that it's inconsistent with both `Length` (existing 1D arrays) and `Shape` (from the multidimensional arrays proposal).

bettinaheim: If we go with size here then we should choose Size for the multidimensional arrays as well. @cgranade

swernli: +1 for keywords that are searchable. While I like the readability of `[10 of “”]` it does make it harder to find online or in code via search.

cgranade: With respect to the 𝑛-d proposal, the term "shape" aligns with Python / NumPy, Rust / ndarray, C# / NumSharp, and C++ / xtensor, but the term "size" is used by Julia, by C++ / Eigen. Meanwhile, the term "length" is used by C# (e.g.: `GetLength(int axis)`). Not content to pick a single term, MATLAB uses a mix of "size" and "shape," so no inspiration there.

While "shape" is more popular, I don't think it's so overwhelmingly so as to preclude moving to "size."

With respect to the immediate syntax question, though, can definitely get behind `["" of size 10]`, `["" of size (3, 4)]` and so forth (still sad about `of`, but I'll get over it, heh). I do want to quickly raise one other possibility inspired by Ruby, though. In particular, if we think of there being a parallel between array comprehension syntax and loop syntax, then `[expr for iter in arrayOrRange]` makes sense as the correspondance to `for (iter in arrayOrRange) { ... } `. We could imagine a new shorthand for `for` loops where we don't care about the iteration variable, following Ruby's `3.times do ... end` syntax:

```
times (10) {
     // this block repeats 10 times
}
```

In the same sense that this loop is short for `for (_ in 1..10) { ... }`, we could have a shorthand array comprehension `[expr times 10]`. The same argument would also hold for `[expr repeat 10]` being a shorthand for `repeat (10) { ... }` if we go that route. As @bettinaheim pointed out offline, this does have the disadvantage of mapping a bit less nicely onto the multidimensional case unless we have a nice syntax for multidimensional comprehensions (could be useful anyway).

swernli: I like the idea of tying the syntax to loops, as it gives a commonality and makes the array generation seem seem like less of outlier. On the topic of shorthand, the `times` and `repeat` with an integer lines up with what I've been imagining for a proposal regarding integer limits on repeat-until-success loops. Then you could write something like
```qsharp
repeat (10) {
    ApplyOp(q);
}
until (M(q) == Zero);
```
And that would run the body either 10 times or until the condition is true, whichever comes first. But that's a separate conversation for another thread...

cgranade: `[Measure([PauliZ repeat 3], register) repeat until M(Head(register)) == One]`

bettinaheim: I was thinking about how that looks like together with comprehension and the multidimensional case. I am a bit worried that that if we have `<expr> size <expr>` it is harder to read and looks like a single operator expression. I think having an equal sign actually helps readability.
Here is what I've got, assuming syntax for multidimensional arrays and array comprehension: 
```qsharp
// all of these create an array of arrays of size 3
[0 size=(3,2)] 
[[x*x for x in 1..2] size=3] 
[[0 size=2], [0 size=3], [0 size=4]]
[[1,2], [3,4,5], [6,7,8,9]]

// all of these create a 2D array of size (3,2)
#[0 size=(3,2)] 
#[[x*x for x in 1..2] size=3] 
#[[0 size=2], [1 size=2], [2 size=2]]
#[[1,2], [3,4], [5,6]]

// all of these create a 3D array of size (4,3,2)
#[0 size=(4,3,2)] 
#[#[[x*y for x in 1..2] for y in 1..3] size=4] 
#[#[[1,2], [3,4], [5,6]], #[[1,2], [3,4], [5,6]], #[[1,2], [3,4], [5,6]], #[[1,2], [3,4], [5,6]]] 
```

anpaz (commented):
My preference would be:
1. Remove `new`
2. Provide the methods proposed by Sarah:
  * `EmptyArray<T>()`
  * `ConstantArray(n, value)`
3. Provide sugar for:
   * `T[]` ==> `EmptyArray<T>()`
   * `T[n]` ==> `ConstantArray(n, Default)`; and have the compiler fail for any `T` that has no default (like `Qubit` and `String`).

Approved/remove-new-array.md
cgranade: > I was thinking about how that looks like together with comprehension and the multidimensional case. I am a bit worried that that if we have `<expr> size <expr>` it is harder to read and looks like a single operator expression. I think having an equal sign actually helps readability.

Would we similarly want an `=` sign in array comprehensions, then, or some other operator to make it clear that comprehensions aren't expressions in their own right?

guenp:
I agree with @anpaz's suggestion

Approved/remove-new-array.md
bettinaheim: Well, comprehensions are more expressions on their own right - they are like ranges in a sense. For to me implies an iteration, size does not.

cgranade:
> My preference would be:
> 
> 1. Remove `new`
> 2. Provide the methods proposed by Sarah:
> 
> * `EmptyArray<T>()`
> * `ConstantArray(n, value)`

Currently, both of these functions already exist in Microsoft.Quantum.Arrays:

- [ConstantArray function](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.arrays.constantarray)
- [EmptyArray function](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.arrays.emptyarray)

> 1. Provide sugar for:
>    
>    * `T[]` ==> `EmptyArray<T>()`

I think following the above suggestion from @samarsha, if we want to annotate types of expressions, we should likely do so as a general feature and not as a one-off for empty arrays. Since types in Q# always follow values in declarations (e.g.: `target : Qubit`), it feels like annotating empty arrays as `[] : 'T` would be more consistent with that convention.

samarsha:
> `T[n]` ==> `ConstantArray(n, Default)`; and have the compiler fail for any T that has no default (like `Qubit` and `String`).

This means that users will need to memorize not only what the default values are for each type, but which types have default values. It saves a few characters when writing, but IMO it's harder to read than explicitly writing out the value you want to use. I think we should optimize for readability and consistency here.

By contrast, if you require array initializers to include the value explicitly, users only need to memorize the initialization syntax. Always including the value also more intuitively generalizes to initializing arrays of non-default values: instead of needing to switch from `Int[n]` for an array of `0`s, to `ConstantArray(n, 1)` for an array of `1`s, they would both be `ConstantArray(n, 0)` and `ConstantArray(n, 1)` (or whatever syntax we end up with).

Approved/remove-new-array.md
bettinaheim: Alright, I think we need to come to a decision, and while I believe it is a good choice to keep the syntax restricted to within rectangular brackets for design reasons, there isn't an objectively best choice for the precise syntax. I am coming back to `[M(q), size=10]` for the following reasons: 
- I see it as a call to a constructor, and the two items within rectangular brackets as arguments. The distinction that it is a call is important since `M(q)` would be evaluated exactly once (also when the size is 0). This is in contrast to `[M(q) for _ in 1..10]` where `M(q)` would indeed be evaluated 10 times. 
- `size` is a full, unique, and short word that can be searched for in our docs.

bettinaheim: Thinking one step further, I believe it may actually be reasonable to support things like 
```qsharp
let buildTuple = (1,2,_); // callable to construct a tuple; the exact type of the callable will be inferred based on the first usage
let populateArray = [_, size=10]; // same here
```
in the future (though that should be worked out separately).

cgranade: Would it be possible to do it the other way, to have syntax that parallels `ConstantArray(_, "foo")` to partially apply the element but not the size? Something like `let bySize = ["foo", size=_]`?

bettinaheim: Yes, I would expect so.

Approved/remove-new-array.md
bettinaheim: Let's support empty array literals (`[]`) as part of implementing this proposal. Type inference based on the first usage is something we should do sooner or later in any case, so let's start with this.

cgranade: 💕 for empty array literals!

guenp:
@cgranade @samarsha I agree that `let myArray = [] : Int` would work great.

Regarding using `Int[4]` instead of `ConstantArray(4, Default)`; for this example specifically, I can see that `let myArray = Int[4]` might create ambiguity with `let myArray = [4]`.

Ideally I would suggest `let myArray = [0] * 4 # same as [0, 0, 0, 0]`, but @samarsha explained above why that won't work for us:

> We would need to choose a different operator name than *. Otherwise, it's not possible to reconcile the type of * used for arrays with the type of * needed to multiply two numbers together. This would make it very difficult to create a type class to represent numeric types. See microsoft/qsharp-compiler#557.

I would be fine coming up with a new operator for this; would something like this work for example?
```qsharp
let myArray = [0] : Int.length(4) # same as [0, 0, 0, 0]
```

alan-geller:
> @cgranade @samarsha I agree that `let myArray = [] : Int` would work great.
> 
> Regarding using `Int[4]` instead of `ConstantArray(4, Default)`; for this example specifically, I can see that `let myArray = Int[4]` might create ambiguity with `let myArray = [4]`.
> 
> Ideally I would suggest `let myArray = [0] * 4 # same as [0, 0, 0, 0]`, but @samarsha explained above why that won't work for us:
> 
> > We would need to choose a different operator name than *. Otherwise, it's not possible to reconcile the type of * used for arrays with the type of * needed to multiply two numbers together. This would make it very difficult to create a type class to represent numeric types. See [microsoft/qsharp-compiler#557](https://github.com/microsoft/qsharp-compiler/issues/557).
> 
> I would be fine coming up with a new operator for this; would something like this work for example?
> 
> ```qsharp
> let myArray = [0] : len(4) # same as [0, 0, 0, 0]
> ```

I don't think this works either. In Q#, it's only type names that come after a `:`, which would imply that `len(4)` is a type name. Now we've got dependent types creeping in, with an implication that `len(4)` and `len(2)` are different types.

I'm not sure why this has problems with type classes. Why can't there be a type class "Multipliable" that contains arrays and numbers, and `*` supports multiplying Multipliables by numbers? `+` will have the same need in order to support array and string concatenation.

guenp:
Thanks, @alan-geller , that makes sense. For completeness, @samarsha  also mentioned the issue of overloading with `let myArray = [0] * 4`;

> The type of * is overloaded to both ('a[], Int) -> 'a[] and (Int, 'a[]) -> 'a[], since Python also supports 10 * [1]. Q# does not support overloading. I would argue that there are good reasons for Q# to never support overloading, but there is a feature request for this: microsoft/qsharp-compiler#29

Aside from whether this would lead to a huge amount of work or not, I'm curious if this would make sense:
```qsharp
let myArray = [0 for i in 0..3] # same as [0, 0, 0, 0]
```

alan-geller:
> Thanks, @alan-geller , that makes sense. For completeness, @samarsha also mentioned the issue of overloading with `let myArray = [0] * 4`;
> 
> > The type of * is overloaded to both ('a[], Int) -> 'a[] and (Int, 'a[]) -> 'a[], since Python also supports 10 * [1]. Q# does not support overloading. I would argue that there are good reasons for Q# to never support overloading, but there is a feature request for this: [microsoft/qsharp-compiler#29](https://github.com/microsoft/qsharp-compiler/issues/29)
> 
> Aside from whether this would lead to a huge amount of work or not, I'm curious if this would make sense:
> 
> ```qsharp
> let myArray = [0 for i in 0..3] # same as [0, 0, 0, 0]
> ```

To me, that makes perfect sense. It's just like [i*i for i in 0..3] to get [0,1,4,9}.

cgranade:
> I'm not sure why this has problems with type classes. Why can't there be a type class "Multipliable" that contains arrays and numbers, and `*` supports multiplying Multipliables by numbers? `+` will have the same need in order to support array and string concatenation.

If I understood right, it was that @samarsha's proposal for a typeclass / concept representing types supporting `*` captured a signature of the form `(*) : ('T, 'T) -> 'T` (that is, a _magma_), but `3 * ["foo"]` yielding `["foo", "foo", "foo"]` looks more like thinking of `*` as the group action of `Int` on `String[]` (in particular, `(n + m) * data` is the same as `n * data + m * data`, so group action axioms are preserved). Using the same operator for the magma and group action typeclasses may be confusing in some contexts, perhaps?

(On a side note, I say magma and not semigroup since there's no way at the type system level to enforce associativity; strictly speaking, `3 * ["foo"]` may fail to be a group action for the same reason, but I think that points to the need for type classes to come with their own unit tests that enforce invariants of those type classes.)

samarsha:
> Why can't there be a type class "Multipliable" that contains arrays and numbers, and `*` supports multiplying Multipliables by numbers?

Sorry for the Haskell, but is this what you mean?

```haskell
class Multipliable a where
    (*) :: Num b => a -> b -> a
    -- or
    (*) :: Num b => b -> a -> a
```

This type seems to have unfortunate implications. For example, if `a ~ Int` and `b ~ Double`, then an `Int` times a `Double` must yield an `Int`. Also, it would allow arrays to be multiplied by `Double`, unless we restrict `Num b` to `Int`, but then how do we define multiplication between two `Double`s?

> `+` will have the same need in order to support array and string concatenation.

I don't think `+` necessarily has the same problem for concatenation if we use a monoid. Both numeric `+` and concatenation `+` would have the same type, `Monoid a => a -> a -> a`.

bettinaheim:
@guenp @alan-geller Since the resolution to the syntax question is a bit buried in the rather long thread above, let me recap here: 
- We will add support for empty array literals (`[]`), no type annotation needed (or supported at this time). 
- We will add the syntax `["", size=5]` for creating arrays with all items instantiated to a given value (`""` in this case). 

The reason for the choice of syntax is the following: 
- The syntax restricted to within rectangular brackets to not limit or negatively impact future capabilities. 
- The construct resembles and should be treated as a constructor call, and the two items within rectangular brackets as the corresponding argument items (while passing argument items by name is not yet supported, it is something we want to add in the future). The distinction that it is effectively a constructor call and dissimilar to the array comprehension syntax is important since M(q) is evaluated exactly once (also when the size is 0). This is in contrast to [M(q) for _ in 1..10] where M(q) would indeed be evaluated 10 times.
- `size` is a full, unique, and short word that can be searched for in our docs.

With this interpretation, we can build on it in the future and potentially fully support partial application also for such built-in syntax constructs (to be worked out in a separate proposal), meaning things like these:
```qsharp
// In all cases, the exact type of the callable will be inferred based on the first usage. 
let buildTuple = (1,2,_); // callable to construct a tuple
let populateArray = [_, size=10]; // callable to construct an array
let createArray = [Zero, size=_]; // callable to construct an array
let multiply = _ * _;  
```
In combination with array comprehension (to be worked out and confirmed in a separate proposal) and multidimensional arrays (a proposal is in progress), we get the following: 
```qsharp
// all of these create an array of arrays of length 3
[0, size=(3,2)] 
[[x*x for x in 1..2], size=3] 
[[0 size=2], [0 size=3], [0 size=4]]
[[1,2], [3,4,5], [6,7,8,9]]

// all of these create a 2D array of shape (3,2)
#[0, size=(3,2)] 
#[[x*x for x in 1..2], size=3] 
#[[0 size=2], [1 size=2], [2 size=2]]
#[[1,2], [3,4], [5,6]]

// all of these create a 3D array of shape (4,3,2)
#[0, size=(4,3,2)] 
#[#[[x*y for x in 1..2] for y in 1..3], size=4] 
#[#[[1,2], [3,4], [5,6]], #[[1,2], [3,4], [5,6]], #[[1,2], [3,4], [5,6]], #[[1,2], [3,4], [5,6]]] 
```

cgranade:
> > Why can't there be a type class "Multipliable" that contains arrays and numbers, and `*` supports multiplying Multipliables by numbers?
> 
> Sorry for the Haskell, but is this what you mean?
> 
> ```haskell
> class Multipliable a where
>     (*) :: Num b => a -> b -> a
>     -- or
>     (*) :: Num b => b -> a -> a
> ```
> 
> This type class seems to have unfortunate implications. For example, if `a ~ Int` and `b ~ Double`, then an `Int` times a `Double` must yield an `Int`. Also, it would allow arrays to be multiplied by `Double`, unless we restrict `Num b` to `Int`, but then how do we define multiplication between two `Double`s?
> 
> > `+` will have the same need in order to support array and string concatenation.
> 
> I don't think `+` necessarily has the same problem for concatenation if we use a monoid. Both numeric `+` and concatenation `+` would have the same type, `Monoid a => a -> a -> a`.

I think we can have both behaviors "just work" if we separate the `*` for multiplicative magmas from the `*` for the left group action of integers on `'T[]`. Concretely, consider the following sequence of type classes and unit test templates:

```
concept 'T is MultiplicativeMagma when {
    function (*)(left : 'T, right : 'T) : 'T;
}

concept 'T is MultiplicativeSemigroup where 'T is MultiplicativeMagma, EquatableTo<'T> when {
    invariant Associative(left : 'T, middle : 'T, right : 'T) : Unit {
        if ((left * middle) * right != left * (middle * right)) { fail ""; }
    }
}

concept 'T is MultiplicativeMonoid where 'T is MultiplicativeSemigroup when {
    function MOne() : 'T;

    invariant LeftIdentity(right : 'T) : Unit {
        if (MOne() * right != right) { fail ""; }
    }

    invariant RightIdentity(left : 'T) : Unit {
         if (left * MOne() != left) { fail ""; }
    }
}

concept 'T is MultiplicativeGroup where 'T is MultiplicativeMonoid when {
    function Reciprocal(self : 'T) : 'T;

    // TODO: invariants for x * (1/x) = 1 and (1/x) * x = 1.
}

concept 'T is AdditiveMagma when {
    function (+)(left : 'T, right : 'T) : 'T;
}

// TODO: other concepts for additive versions of semigroup, monoid, and group.

concept 'T is LeftActsOn<'U> where 'T is MultiplicativeGroup, 'U is EquatableTo<'U> {
    function (**)(left : 'T, right : 'U) : 'U;

     invariant Associative(g : 'T, h : 'T, x : 'U) : Unit {
         if ((g * h) ** x != g ** (h ** x)) { fail ""; }
     }

     invariant LeftIdentity(x : 'U) : Unit {
         if (MOne() ** x != x) { fail ""; }
     }
}
```

samarsha:
Yes, if the operator names are different, then I don't think there's any issue. My concern was only with re-using `*` for both cases.

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,149 @@
...
+### Repeated constant arrays
+
+The semantics of `[x, size = n]` are identical to `new T[n]`, except that the initial value is now given by `x`:
+
+1. `n` may be an integer literal or a variable of integer type.
+2. If `n` is negative, a runtime error occurs.
+
+```qsharp
+// Create an array of length 10 of type Bool[] where every value is false.
+let bools = [false, size = 10];
```
samarsha, line 111: I'm still concerned about the inconsistent terminology: to create an array of *length* n, use the *size* option...

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+   Its use with default-initialized arrays is inconsistent with other type constructors.
+2. It requires every type to have a default value with which to initialize each item in the array.
+
+### Default values
+
+It is not possible to define reasonable default values of types like `Qubit` and `a -> b`.
+Their current default values are invalid and will trigger an error if they are used.
+But the syntax is the same both for types with valid default values, and those without: while `new Int[3]` is a well-defined value whose items can be used safely, `new Qubit[3]` creates a trap that will crash the program if any item in the array is used before being set to a valid value first.
+
+This assumption is also invalid for uninhabited types such as `Void`.
```
bettinaheim, line 31:
```suggestion
This assumption is also invalid for uninhabited types such as `Void` in other languages.
```

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+If Q# requires that every type has a default value, then uninhabited types are not possible to express properly.
+This can even cause subtle bugs in the soundness of a type system, as demonstrated by [this bug in Java's generics](https://hackernoon.com/java-is-unsound-28c84cb2b3f) that happened because of the existence of a value (specifically `null`) for a type that should have been uninhabited.
+
+Removing default-initialized array constructors, and the requirement that every type has a default value, resolves these problems.
+
+## Enhancing array literals
+
+With the removal of default-initialized array constructors, alternatives are needed to create empty arrays and arrays with a repeated initial value.
+The alternative for `new T[0]` is `[]`.
+The alternative for `new T[n]` where `n > 0` is `[x, size = n]`, where `x` is the initial value of each item.
```
bettinaheim, line 41: n = 0 should be ok here as well. 
```suggestion
The alternative for `new T[n]` where `n >= 0` is `[x, size = n]`, where `x` is the initial value of each item.
```

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+let empty = new Int[0];
+
+// Create an array of length 10 of type Bool[] where every value is false.
+let bools = new Bool[10];
+```
+
+## Proposed Modification
+
+The syntax above will be deprecated and removed.
+The replacement for each case is different.
```
bettinaheim, line 60:
```suggestion
```

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+    let empty2 = [];
+
+    // While empty3 is used, its usage does not provide any information about its concrete type.
+    // This is a compilation error.
+    let empty3 = [];
+    Message($"Empty array: {empty3}");
+
+    // In cases where the type of [] cannot be inferred, Microsoft.Quantum.Arrays.EmptyArray can be used instead.
+    // The following code is valid.
+    let empty4 = EmptyArray<Int>();
```
bettinaheim, line 99:
```suggestion
    let empty4 = EmptyArray<Int>(); // or alternatively [0, size=0]
```

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+    Message($"Empty array: {empty4}");
+}
+```
+
+### Repeated constant arrays
+
+The semantics of `[x, size = n]` are identical to `new T[n]`, except that the initial value is now given by `x`:
+
+1. `n` may be an integer literal or a variable of integer type.
```
bettinaheim, line 108: Why the restriction here? I think `n` can be any integer expression.

Approved/enhanced-array-literals.md
bettinaheim: Well, when we introduce multi-dimensional arrays it would make sense to introduce a Size function that can be used also for 1D arrays. I think using Length for multidimensional arrays really doesn't work, and it would seem odd to not support Size for the 1D case. @cgranade Does supporting a Size function for arrays of any dimension make sense to you and if so, can we introduce support for it (for the 1D case) before we introduce multidimensional arrays? I would suggest to keep Length, which would mean that both Length and Size are valid and return the same for the 1D case.

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+```qsharp
+use qs = Qubit[n];
+```
+
+For consistency, it may make sense to change this syntax as well.
+The [proposal for allocatable types and generalization of initializers](https://github.com/microsoft/qsharp-language/pull/41) is addressing this.
+
+## Impact on Existing Mechanisms
+
+The [`Default`](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.core.default) function depends on default-initialized array constructors, and it will also be deprecated.
```
bettinaheim, line 153: @cgranade for awareness

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+
+let g = [_, size = 3];
+// g : 'a -> 'a[]
+// g("foo") == ["foo", "foo", "foo"]
+
+let h = [_, "middle", _];
+// h : (String, String) -> String[]
+// h("first", "last") == ["first", "middle", "last"]
+```
+
```
bettinaheim, line 187: The same would be supported for other built-in functions such as addition:
```qsharp
let f = 1 + _;
// f : Int -> Int
// f(3) == 4

let g = _ + _
// g : 'T -> 'T where the type of 'T will be determined by the first usage
// g(1.,1.) == 2. 
```

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+For consistency, it may make sense to change this syntax as well.
+The [proposal for allocatable types and generalization of initializers](https://github.com/microsoft/qsharp-language/pull/41) is addressing this.
+
+## Impact on Existing Mechanisms
+
+The [`Default`](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.core.default) function depends on default-initialized array constructors, and it will also be deprecated.
+
+## Anticipated Interactions with Future Modifications
```
bettinaheim, line 155: It would be good to add a paragraph on both multidimensional arrays (that is actually quite an important factor), and also the possibility of array comprehension.

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,209 @@
...
+For repeated constant arrays, the syntax `[x, size = n]` is not as obvious.
+Many languages use a function to provide this functionality instead of built-in syntax.
+However, we were not able to come to a consensus on how the `ConstantArray` function should be made available.
+There are unresolved questions, such as:
+
+1. Should `ConstantArray` be added to `Microsoft.Quantum.Core` so that it can be used without an open directive?
+2. Should `ConstantArray` be part of a more general `Array` module that is in `Microsoft.Quantum.Core`, so that it can be used like `Array.Constant`?
+3. Should modules be a language feature in Q# or are open-as directives enough to emulate them?
+
+We defer answering these questions for now, instead adding new syntax for this case.
```
bettinaheim, line 205: Saying we added new syntax because we couldn't come to a consensus is accurate. Given the inconveniences and downsides of each possible ways in which we could expose this via a library function and given how common array constructions are, it is justified to introduce new syntax. You can list the downsides of the exposing it via a library (copy-pasting from the rather long thread), or you can leave it at that.

Approved/enhanced-array-literals.md
samarsha: Yes, that's correct, just bad phrasing on my part.

Approved/enhanced-array-literals.md
samarsha: I disagree that the new syntax is justified when compared with an alternative function API that we could agree on. The thread is very long, but I don't remember any downsides of using a function other than needing to decide what to call it and where to put it. I remember some arguments about wanting to avoid duplicating `ConstantArray`, `ConstantSeq`, etc. if Q# has multiple collection types, but these same arguments apply to `[x, size = n]`, which is also array-specific.

bettinaheim: There are downsides to every possibility that are discussed in detail in the thread. I understand your assessment is that there are significant downsides to introducing new syntax that outweight the benefits of providing an alternative to a library call (which you can also use). I very much appreciate your thoughts and discussions about it with you (hence the length of the thread above). Given that we take care that the introduced syntax does not negatively impact future features and ties in well with anticipated features, I believe the best path forward is nonetheless to provide this alternative.

Approved/enhanced-array-literals.md
samarsha: Is this relevant to the proposal? It seems like partial operator application doesn't interact with array literals.

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,251 @@
...
+
+In the future, the type inference used for empty array literals may be extended to all expressions, allowing calls to polymorphic functions where not all type variables are determined by their arguments.
+
+### Partial application
+
+We may want to consider supporting partial application for array literals in the future:
+
+```qsharp
+let f = ["foo", size = _];
+// f : Int -> String[]
```
samarsha, line 179: After thinking about this some more, I don't think the assertion that `f : Int -> String[]` is accurate in general. I think partial application of the `size` parameter is problematic. See the Raised Concerns section for more about this.

Approved/enhanced-array-literals.md
```diff
@@ -0,0 +1,251 @@
...
+    let empty5 = EmptyArray<Int>();
+    Message($"Empty Int arrays: {empty4}, {empty5}");
+}
+```
+
+### Repeated constant arrays
+
+The semantics of `[x, size = n]` are identical to `new T[n]`, except that the initial value is now given by `x`:
+
+1. `n` is an expression of type `Int`.
```
samarsha, line 111: This sentence will not be accurate in the future, since we expect to be able to provide an `Int` tuple of any arity to the `size` parameter. This type cannot be expressed in Q#'s type system, which concerns me.

samarsha: I think one way to fix this is to forbid things like `[0, size = (3, 2)]` to construct nested arrays. If we restrict this to multidimensional arrays, then the type of `size` is determined by the number of `#`s, so `#[0, size = (n, n)]`, `##[0, size = (n, n, n)]`, etc.

samarsha: To construct a nested array, I recommend the more verbose but more explicit syntax `[[0, size = _], size = _]`.

samarsha: @bettinaheim, can you clarify whether you want to support these examples from your comment here: https://github.com/microsoft/qsharp-language/pull/48#issuecomment-732601443

* `[0, size = (3, 2)]`
* `#[0, size = (4, 3, 2)]`

Or if we can enforce stricter typing like in my examples above?

* `[[0, size = _], size = _]`
*  `##[0, size = (4, 3, 2)]`

Approved/enhanced-array-literals.md
samarsha: Updated the proposal to clarify that a fixed type for `size` should be used.

Approved/enhanced-array-literals.md
samarsha: Updated the proposal to recommend against overloading on tuple arity for `size`, and prefer syntax with unambiguous types.

bettinaheim: Thanks for bringing this up. That it should always be possible to translate the introduced construct to a function call is a very good point and I agree with ensuring that we limit its usage in the multidimensional case to ensure that.

bettinaheim (dismissed):
Thank you for working through this! This has turned into a great proposal and has spawned many useful conversations!

bettinaheim (approved):

PR #54, opened 2020-11-18
Added specs for dealing with bad data and other errors

### LINKED ISSUE #30: QIR: consider specifying the expected behaviour in error cases such as out of bounds access to arrays
Currently, few of the __quantum__rt__* methods prescribe what should happen in case of an error. For example:
- using a range with zero step,
- getting an out of bounds element from an array,
- projecting a 1D array,
- concatenating non-1D arrays,
etc.

Leaving those cases unspecified means undefined behaviour, specific to each vendor implementation of the runtime. While it's a possible choice, it might be worth to consider being more prescriptive. And if we choose to go with UB, the spec should probably explicitly state so in the overview section.

alan-geller:
Resolves #30


Specifications/QIR/Data-Types.md
```diff
@@ -264,6 +287,8 @@ an array should cause an immediate runtime failure.
 This applies to slicing and projection operations as well as to element access.
 When validating indices for slicing, only indices that are actually part of the
 resolved range should be considered.
+When projecting, an attempt to project a one-dimensional array on its only dimension
```
bettinaheim, line 290: Unless I am misunderstanding what you mean by projecting on its only dimension: why shouldn't the projection on its only dimension return the array itself?

Specifications/QIR/Callables.md
```diff
@@ -3,6 +3,12 @@
 We use to term _callable_ to mean a subroutine in the source language.
 Different source languages use different names for this concept.
 
+### Runtime Failure
+
+There are several error conditions that are specified as causing a runtime failure.
+The `quantum__rt__fail` function is the mechanism to use to cause a runtime failure;
```
bettinaheim, line 9: Reading through what failures we are talking about, that seems reasonable. At some point in the future, there may be a need for having a "fail" instruction on the target machine in order for it to give more meaningful information about the state at failure, but that seems to be a different issue than the failures needed here.

Specifications/QIR/Data-Types.md
```diff
@@ -99,6 +109,8 @@ simple types:
 | __quantum__rt__result_reference   | `void(%Result*)`         | Increments the reference count of a Result pointer. |
 | __quantum__rt__result_unreference | `void(%Result*)`         | Decrements the reference count of a Result pointer and releases the result if appropriate. |
 
+For `equal`, if a `%Result*` parameter is null, a runtime failure should occur.
```
irinayat-MS, line 112: I'd suggest to use full function names all the time, as it makes the specifications more searchable.

Specifications/QIR/Data-Types.md
irinayat-MS: Projection must decrease array's dimensionality and we don't support 0-dimentional arrays. Maybe we could treat scalars as such but is it worth the complexity?

Specifications/QIR/Data-Types.md
```diff
@@ -272,7 +297,7 @@ arrays:
 |----------------------------------|--------------------------------------|-------------|
 | __quantum__rt__array_create_1d   | `%Array* void(i32, i64)`             | Creates a new 1-dimensional array. The `i32` is the size of each element in bytes. The `i64` is the length of the array. The bytes of the new array should be set to zero. If the length is zero, the result should be an empty 1-dimensional array. |
 | __quantum__rt__array_copy        | `%Array*(%Array*)`                   | Returns a new array which is a copy of the passed-in `%Array*`. |
-| __quantum__rt__array_concatenate | `%Array*(%Array*, %Array*)`          | Returns a new array which is the concatenation of the two passed-in arrays. |
+| __quantum__rt__array_concatenate | `%Array*(%Array*, %Array*)`          | Returns a new array which is the concatenation of the two passed-in one-dimensional arrays. If either array is not one-dimensional, then a runtime failure should occur. |
```
irinayat-MS, line 300: The arrays also must contain elements of the same size.
Also, is it allowed to concatenate arrays of qubits? If so, how the arrays/qubits should be released?

Specifications/QIR/Data-Types.md
bettinaheim: I think I am just tripping up on the formulation; thinking geometrically, to remove one dimension, you project onto the perpendicular hyperspace. Saying that it is projected onto its dimension sounds like it is projected on the axis it is already on in the 1D case.

Specifications/QIR/Data-Types.md
bettinaheim: Yes, arrays of qubits can be concatenated. I'd like to understand the question around how they should be released better. What is the relation between array concatenation and releasing?

irinayat-MS: When creating a new array of qubits, one must then release it with `__quantum__rt__qubit_release_array`. The question is, should this call release the qubits the array contains or is it the client's responsibility to release the qubits manually and only after that release the array itself? So far I've been assuming the former (otherwise, why provide a separate array alloc/release for arrays of qubits?). In either case, the description of `__quantum__rt__qubit_release_array` should probably be clarified.

irinayat-MS: Thinking some more about it, `__quantum__rt__qubit_release_array` seems to be just syntactic sugar over 1) release all qubits it contains, 2) release the array -- given how easy it is to cause double release of qubits by using it (if we allow arrays of qubits to be copied, concatenated, etc.) -- should it even be provided?

Arguably, `__quantum__rt__qubit_allocate_array` is also just a shortcut, but it's safe so I don't have strong feelings against providing it.

Specifications/QIR/Quantum-Runtime.md
```diff
@@ -38,6 +44,9 @@ The code that borrows the qubits guarantees that the state of the qubits when
 returned is identical, including entanglement, to their state when borrowed.
 It is always acceptable to satisfy `borrow` by allocating new qubits.
 
+Passing a null `%Qubit*` or `%Array*` to any of these functions should
```
irinayat-MS, line 47: What about passing a pointer to a non-qubit array to `__quantum__rt__qubit_release_array`? (Per my other comment, maybe, reconsider providing this API at all?)

Specifications/QIR/Data-Types.md
```diff
@@ -294,5 +319,8 @@ There are special runtime functions defined for allocating or releasing an
 array of qubits.
 See [here](Quantum-Runtime.md#qubit-management-functions) for these functions.
 
+For all of these functions other than `reference` or `unreference`, if an
```
irinayat-MS, line 322: What should happen if ref/unref on an array of qubits? What if the ref-count is > 1 but release_qubit_array is called? Basically, does an array of qubits own the qubits it contains or not? I'd propose it does _not_ and is subject to all array operations, leaving it to the clients to ensure they don't concatenate arrays of mismatched types, and that qubits must be released independent of the array's memory management.

Specifications/QIR/Data-Types.md
alan-geller: Maybe I should clean up the wording; geometrically you're always projecting onto the perpendicular hyperspace of the selected dimension. Maybe projecting "along" rather than "on"?

Specifications/QIR/Data-Types.md
alan-geller: I think the rule is that any array that is created with `__quantum__rt__qubit_allocate_array` must be released with `__quantum__rt__qubit_release_array` and no array created in any other way may be released with `__quantum__rt__qubit_release_array`. An array of qubits you create by concatenating two existing arrays of qubits should be unreferenced like any other array.

I'm reluctant to drop `__quantum__rt__qubit_release_array` because replacing it with an LLVM loop will make the QIR even more verbose than it already is. Also, it will encourage people to think that it's OK to allocate an array of qubits and then release them one by one at different times; the intent of allocating and releasing an array is that you're treating the qubits as a group (a quantum register), not as just a pool of random qubits.

It is certainly true that the current text doesn't make this clear. If you're both OK with what I just wrote, I'll add it to the text.

irinayat-MS: To me, it feels counter-intuitive that concatenating two arrays of qubits creates a new array that is managed in a different way than the original arrays. The crossing between arrays of qubits and arrays of other types is already precarious. Would it make sense to have a separate type for arrays of qubits that owns its content (then concatenate, copy, etc. could take arrays of qubits but return non-managing array of `%Array*`)?

Re treating an array of qubits as a register: I'm fine with this, but what it would mean then to copy a register or to concatenate two registers?

bettinaheim: I think making a type distinction is likely to cause trouble. A thing that might be ok is if an array has a flag indicating just that. Even so, I am doubtful it is worth the hassle. To me, this has little to do with the array - allocation and deallocation is not associated with a specific data type. Instead, if/once we have initializers that e.g. allocate user defined types, then we might want to generate a method to allocate that type (calling into __quantum__rt__qubit_allocate_ / __quantum__rt__qubit_allocate_array as needed) and the corresponding deallocate method with it. 
I agree with keeping __quantum__rt__qubit_release_array.

Specifications/QIR/Data-Types.md
bettinaheim: Yes, I think that makes sense.

Specifications/QIR/Data-Types.md
alan-geller: Done.

Specifications/QIR/Data-Types.md
```diff
@@ -18,10 +24,12 @@ All types follow the same pattern:
   count of an instance and an `_unreference` routine that decrements the
   reference count.
 - The `_unreference` routine will release the instance if the reference count
-  is decremented to zero.
-- The `_unreference` routine should accept a null instance pointer and simply
-  ignore the call if the pointer is null. This allows us to avoid null checks
-  strewn through the QIR, with the attendant plethors of LLVM basic blocks.
+  is decremented to zero. It should ignore the call if the reference count is
+  already zero or less than zero.
```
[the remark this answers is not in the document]
alan-geller: I think this is made obsolete by Bettina's PR containing the ref count and alias count updates.

Specifications/QIR/Data-Types.md
alan-geller: I added the requirement that the arrays have the same size. The rest is covered in Bettina's coming PR.

Specifications/QIR/Data-Types.md
alan-geller: Addressed by Bettina's coming PR.

bettinaheim (approved):

PR #55, opened 2020-11-18
Replaced variadic runtime functions with simple functions

### LINKED ISSUE #37: QIR: Consider avoiding variadic functions in the runtime
As far as I can tell, there is no platform-agnostic way to implement variadic functions at the IR level and it's also error-prone, which creates a burden on the runtime implementers.

Instead of requiring:
```
%Array* __quantum__rt__array_create(i32, i32, ...)
i8* __quantum__rt__array_get_element_ptr(%Array*, ...)
```
consider similar to 1D specialization for 2D arrays (the arguments will be still passed via registers) and a generic buffer-based signature for larger arrays:
```
%Array* __quantum__rt__array_create_2d(i32, i64, i64)
i8* __quantum__rt__array_get_element_ptr_2d(%Array*, i64, i64)
%Array* __quantum__rt__array_create(i32, i32 %buffer_length, i64* %buffer)
i8* __quantum__rt__array_get_element_ptr(%Array*, i32 %buffer_length, i64* %buffer)
```
Notes: 
- even though the buffer size should match dimensionality of the array, it makes the signatures more consistent/safe to provide it explicitly in array_get_element_ptr;
- for the size-specific functions a 3d one will also still fit into registers (in the common C calling conventions)

alan-geller:
Resolves #37

### DIFF
--- a/Specifications/QIR/Data-Types.md
+++ b/Specifications/QIR/Data-Types.md
@@ -283,9 +283,11 @@ The following utility functions are provided if multidimensional array support i
 
 | Function                         | Signature                            | Description |
 |----------------------------------|--------------------------------------|-------------|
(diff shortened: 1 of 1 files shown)


irinayat-MS (approved):
Looks great, thank you!

PR #56, opened 2020-11-23
Adding a script through GitHub actions to sync with the docs repo

geduardo:
#### Files 

There are 4 relevant files:

- `uid-map.csv`: this is a `*.csv` file containing the map between paths and uids that Brad created. This file can be hosted anywhere as long the link is accessible. I recommend storing this file in the same folder as the stub files, so any modification or addition can be easily reflected in the map.

- `replace_links.yml`: a `*.yml` file containing the workflow for the GitHub action that is described below. This file is in `.github/workflows`.

- `setup.py`: python file to install the Click command-line application.

- `replace_links.py`: python script that imports the map from the url where the .csv file is hosted and replaces the files in a specific path.

#### Workflow

First, there are at least two different branches, `main` and a copy of main where we are going to host the uid version of the docs (`⭐Docs`). The GitHub Actions workflow does the following:

1. Opens a Windows console and checkouts to the repository.
1. Resets the `⭐Docs` branch to match `main`.
1. Installs the script and runs it through a click command.
1. Commits and pushes the changes to `⭐Docs`.

NOTE: the link for the .csv file in the .yml file should be updated when the relevant PR is merged in the other repository.


.github/workflows/replace_links.yml
```diff
@@ -0,0 +1,34 @@
...
+        fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
+    - name: Commit changes
+      run: |
+        git config --local user.email "action@github.com"
+        git config --local user.name "GitHub Action"
+        $ThisCommit = "${{ github.sha }}"
+        git checkout ⭐Docs
+        git reset --hard $ThisCommit
+        pip install ./sync-script
+        replace_links https://raw.githubusercontent.com/geduardo/quantum-docs-pr/adding-file-script/articles/user-guide/language/uid-map.csv ./Specifications/Language/
```
cgranade, line 23: This should absolutely work, but you probably shouldn't need to use `pip install` on `sync-script`; should be able to run in-place once you install `click` itself.

cgranade, line 23: Fairly minor, but it may work better to check out the `quantum-docs-pr` repo itself rather than relying on `raw.githubusercontent.com`.

cgranade (commented):
This looks pretty reasonable to me, no objections on my end but it'd be good to wait for @bettinaheim to approve. We'll need to test in-place once it gets to `main`, but I think this should likely work as-is.

.github/workflows/replace_links.yml
geduardo: Do you mean this?
```suggestion
        pip install click
        python ./sync-script/replace_links.py https://raw.githubusercontent.com/geduardo/quantum-docs-pr/adding-file-script/articles/user-guide/language/uid-map.csv ./Specifications/Language/
```

And get rid of setup.py?

geduardo: Yeah, I considered that, but since it's only a single file and there are different working repos that will store the file in the process, I considered that the link is easier to avoid headaches with the authentication...

Also, I think I would need to clone the full repo and it would take longer for the script to run, right?

sync-script/README.md
```diff
@@ -0,0 +1,26 @@
+## Script to adapt the documents for the official docs page
```
bettinaheim, line 1: Can we please have these files under Specifications/sync-script instead?

geduardo: Yes, I'll change that

cgranade (approved):

PR #57, opened 2020-11-24
fix issues with multiple H1 headings and insecure http:

Bradben:
FIxing build warnings for the sync-to-docs process

### DIFF
--- a/Specifications/Language/3_Expressions/ComparativeExpressions.md
+++ b/Specifications/Language/3_Expressions/ComparativeExpressions.md
@@ -22,7 +22,7 @@ For instance, the following comparison evaluates to `false` due to rounding erro
 >```
 >for some globally declared function `Bar`. The first line defines a new anonymous function that takes an argument `x` and invokes a function `Bar` with it and assigns it to the variable `f1`. The second line assigns the function `Bar` to `f2`. Since invoking `f1` and invoking `f2` will do the same thing, it should be possible to replace those with each other without changing the behavior of the program. This wouldn't be the case if the equality comparison for functions was supported and `f1 == f2` evaluates to `false`. If conversely `f1 == f2` were to evaluate to `true`, then this leads to the question of determining whether two callable will have the same side effects and evaluate to the same value for all inputs, which is not possible to reliably determine. Therefore, if we would like to be able to replace `f1` with `f2`, we can't allow equality comparisons for callables.  
 
-# Quantitative Comparison
+## Quantitative Comparison
 
 The operators less-than (`<`), less-than-or-equal (`<=`), greater-than (`>`), and greater-than-or-equal (`>=`) define quantitative comparisons. They can only be applied to data types that support such comparisons; these are the same data types that can also support [arithmetic expressions](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/3_Expressions/ArithmeticExpressions.md#arithmetic-expressions). 
 
--- a/Specifications/Language/3_Expressions/ComparativeExpressions.md
+++ b/Specifications/Language/3_Expressions/ComparativeExpressions.md
@@ -1,4 +1,6 @@
-# Equality Comparison
+# Comparative Expressions
+
+## Equality Comparison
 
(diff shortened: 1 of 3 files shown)


Specifications/Language/3_Expressions/ValueLiterals.md
```diff
@@ -147,7 +147,7 @@ No literals exist for values of [operation type](https://github.com/microsoft/qs
 No literals exist for values of [function type](https://github.com/microsoft/qsharp-language/tree/main/Specifications/Language/4_TypeSystem#available-types); functions have to be [declared](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/1_ProgramStructure/3_CallableDeclarations.md#callable-declarations) on a global scope and new functions can be constructed locally using [partial application](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/3_Expressions/PartialApplication.md#partial-application).
 
 
-# Default Values
+## Default Values
```
cgranade, line 150: This puts "default values" on the same level as the different kinds of literals; we may need to demote L135, L141, and so forth to H3 lines to make this change work. On the other hand, default values may be going away soon enough anyway (#48)...

Specifications/Language/README.md
```diff
@@ -1,6 +1,6 @@
 # Q# Language
 
-Q# is part of Microsoft's [Quantum Development Kit](http://www.microsoft.com/quantum), and comes with rich IDE support and tools for program visualization and analysis.
+Q# is part of Microsoft's [Quantum Development Kit](https://www.microsoft.com/quantum), and comes with rich IDE support and tools for program visualization and analysis.
```
cgranade, line 3: Some things like http-vs-https we could check with a markdownlint pass on this repo directly; thoughts, @bettinaheim?

Specifications/Language/3_Expressions/ValueLiterals.md
Bradben: I agree - this was a change to avoid a build warning in our repo. The Q#-lang files still need a structural edit pass (for example, this topic would benefit from some intro material and displaying the literal types in a table).

bettinaheim: The literal types are displayed in a table under the type system chapter. The introduction for this is the readme for the expressions. What exactly is generating a warning in the build? Is it enforcing that there is exactly one H1 header at the top of each file?

Specifications/Language/README.md
bettinaheim: Sure. Would that generate a warning, or be a failing check on a PR? Does that pass do anything else?

Specifications/Language/3_Expressions/ComparativeExpressions.md
```diff
@@ -22,7 +22,7 @@ For instance, the following comparison evaluates to `false` due to rounding erro
 >```
 >for some globally declared function `Bar`. The first line defines a new anonymous function that takes an argument `x` and invokes a function `Bar` with it and assigns it to the variable `f1`. The second line assigns the function `Bar` to `f2`. Since invoking `f1` and invoking `f2` will do the same thing, it should be possible to replace those with each other without changing the behavior of the program. This wouldn't be the case if the equality comparison for functions was supported and `f1 == f2` evaluates to `false`. If conversely `f1 == f2` were to evaluate to `true`, then this leads to the question of determining whether two callable will have the same side effects and evaluate to the same value for all inputs, which is not possible to reliably determine. Therefore, if we would like to be able to replace `f1` with `f2`, we can't allow equality comparisons for callables.  
 
-# Quantitative Comparison
+## Quantitative Comparison
```
bettinaheim, line 25: This is also not the correct hierarchy. If we need exactly one H1 header, then adding a header 
# Comparative Expressions
and making both of the subheaders H2 would make sense.

bettinaheim (changes requested):
The headers for comparative expressions should be adapted.

Specifications/Language/3_Expressions/ValueLiterals.md
Bradben: Yes, only one H1 anywhere in the file.

bettinaheim: Ok, leaving the Default Values header here as H2 is fine - it will go away shortly in any case. But updating the ComparativeExpressions.md per my suggestion would be good.

geduardo:
Can we merge this @cgranade @bettinaheim ?

bettinaheim (approved):

PR #67, opened 2020-12-23
QIR specification update 1

bettinaheim:
This is the work in progress to update the QIR specification to 
- simplify how tuples, user defined types and arrays are represented in QIR, and to 
- add the distinction between reference and access counting and detail the corresponding runtime functions and how to use them


Specifications/QIR/Data-Types.md
```diff
@@ -2,15 +2,11 @@
 
 We define LLVM representations for a variety of classical and quantum data types.
 
-In most cases, we represent complex types as pointers to opaque LLVM structure
-types.
-This allows each target to provide a structure definition appropriate for that
-target.
+Representing the types used for qubits and measurement results as pointers to opaque LLVM structure types allows each target to provide a structure definition appropriate for that target.
```
alan-geller, line 5:
```suggestion
Representing the types used for qubits and measurement results as pointers to
opaque LLVM structure types allows each target to provide a structure definition
appropriate for that target.
```

A minor point: keeping the line length limited makes it easier to read diffs (PRs).

Specifications/QIR/Data-Types.md
```diff
@@ -2,15 +2,11 @@
...
 
-### Reference Counting
+### Reference and Access Counting
 
-In QIR, all types represented as pointers, other than qubits, are reference-counted.
-All types follow the same pattern:
+QIR specifies a set of runtime functions for types that are represented as pointers that may be used by the language specific compiler to expose them as immutable types in the language.
```
alan-geller, line 9:
```suggestion
QIR specifies a set of runtime functions for types that are represented as pointers that may be used by the language-specific compiler to expose them as immutable types in the language. 
```

Specifications/QIR/Data-Types.md
```diff
@@ -170,49 +166,36 @@ big integers.
...
-
-Tuple handles are simple pointers to the underlying data structure.
-When passed to a callable function, tuples should always be passed by reference,
-even if the source language treats them as immutable.
-
-In some situations, a pointer to a tuple of unknown or variable type may be required.
-To satisfy LLVM's strong typing in such a situation, tuple pointers are passed as
-pointers to the initial tuple header field; that is, as a `%TupleHeader*`.
-These pointers should be cast to pointers to the correct data structures by the
+When passed to a callable function, tuples are passed as a pointer to an opaque LLVM structure, `%Tuple`. The pointer is expected to point to the contained data such that it can be cast to the correct data structures by the
```
alan-geller, line 172: In retrospect, this wording ("callable function") is ambiguous. I think it would be better to say "when passed to the callable_invoke runtime function..." or something like that, to be painfully precise.

Specifications/QIR/Data-Types.md
```diff
@@ -170,49 +166,36 @@ big integers.
...
 [below](#callable-values-and-wrapper-functions).
 
-Many languages provide immutable tuples, along with operators that allow a modified
-copy of an existing tuple to be created.
-In QIR, this is represented by creating a new copy of the existing tuple and then
-modifying the newly-created tuple in place.
-If the compiler knows that the existing tuple is not used after the modification,
-it is possible to avoid the copy and modify the existing tuple in place.
+Many languages provide immutable tuples, along with operators that allow a modified copy of an existing tuple to be created.
+QIR permits to efficiently support this by requiring the runtime to track and be able to access the following given a `%Tuple*`:
```
alan-geller, line 179:
```suggestion
QIR supports this by requiring the runtime to track and be able to access the following given a `%Tuple*`:
```

Specifications/QIR/Data-Types.md
```diff
@@ -170,49 +166,36 @@ big integers.
...
+- The size of the tuple in number of bytes
```
alan-geller, line 180:
```suggestion
- The size of the tuple in of bytes
```

Specifications/QIR/Data-Types.md
```diff
@@ -170,49 +166,36 @@ big integers.
...
+- The user count indicating how many handles to the tuple exist in the source code
```
alan-geller, line 181: I'm thinking "alias" might be a better word than "user".

Specifications/QIR/Data-Types.md
```diff
@@ -170,49 +166,36 @@ big integers.
...
 
-The following utility functions are provided by the classical runtime to support
-tuples and user-defined types:
+The language specific compiler is responsible for injecting calls to increase and decrease the user count as needed, as well as to accurately reflect when references to the LLVM structure representing a tuple are created and removed. 
+See the section [above](#reference-and-access-counting) regarding the distinction between access and reference counting. 
+
+In the case where the source language exposes tuples as value types rather than reference types, the language specific compiler is expected to request the necessary copies prior to modifying the tuple in place.
```
alan-geller, line 186:
```suggestion
In the case where the source language treats tuples as immutable values, the language-specific compiler is expected to request the necessary copies prior to modifying the tuple in place. 
```

Specifications/QIR/Data-Types.md
```diff
@@ -170,49 +166,36 @@ big integers.
...
+- The size of the tuple in number of bytes
+- The user count indicating how many handles to the tuple exist in the source code
 
-The following utility functions are provided by the classical runtime to support
-tuples and user-defined types:
+The language specific compiler is responsible for injecting calls to increase and decrease the user count as needed, as well as to accurately reflect when references to the LLVM structure representing a tuple are created and removed. 
+See the section [above](#reference-and-access-counting) regarding the distinction between access and reference counting. 
+
+In the case where the source language exposes tuples as value types rather than reference types, the language specific compiler is expected to request the necessary copies prior to modifying the tuple in place. 
+This is done by invoking the runtime function `__quantum__rt__tuple_copy` to create a byte-by-byte copy of a tuple. Unless the copying is forced via the second argument, the runtime may omit copying the value and instead simply return a pointer to the given argument if the user count is 0 and it is hence save to modify the tuple in place.
```
alan-geller, line 187:
```suggestion
This is done by invoking the runtime function `__quantum__rt__tuple_copy` to create a byte-by-byte copy of a tuple. Unless the copying is forced via the second argument, the runtime may omit copying the value and instead simply return a pointer to the given argument if the user count is 0 and it is therefore safe to modify the tuple in place.
```

Also, why do we need a "forced" copy? For that case, why wouldn't the compiler just generate the proper sequence of loads and stores (and user/ref count updates for inner pointers)?

alan-geller (approved):
Mostly purely editorial changes. I'm not sure the term "user count" is very descriptive; might "alias count" or something else be better?

Specifications/QIR/Data-Types.md
```diff
@@ -2,15 +2,11 @@
...
+Representing the types used for qubits and measurement results as pointers to opaque LLVM structure types allows each target to provide a structure definition appropriate for that target.
 
-### Reference Counting
+### Reference and Access Counting
 
-In QIR, all types represented as pointers, other than qubits, are reference-counted.
-All types follow the same pattern:
+QIR specifies a set of runtime functions for types that are represented as pointers that may be used by the language specific compiler to expose them as immutable types in the language. 
 
 - Runtime routines that create a new instance always initialize the instance
```
irinayat-MS, line 11: How the user count should be initialized?

irinayat-MS (approved):

Specifications/QIR/Data-Types.md
bettinaheim: I'll add a sentence that the user count is supposed to be initialized to zero at the corresponding create runtime function.

Specifications/QIR/Data-Types.md
irinayat-MS: I like the idea! It also makes sense for the default "0" value of the count: there is a user, but no aliases.

Specifications/QIR/Data-Types.md
irinayat-MS: There will be legitimate cases when a client might want to copy even in absence of aliases, no?

Specifications/QIR/Data-Types.md
```diff
@@ -170,49 +168,36 @@ big integers.
...
 
 | Function                         | Signature             | Description |
 |----------------------------------|-----------------------|-------------|
-| __quantum__rt__tuple_create      | `%TupleHeader*(i64)`  | Allocates space for a tuple requiring the given number of bytes and sets the reference count to 1. |
-| __quantum__rt__tuple_reference   | `void(%TupleHeader*)` | Indicates that a new reference has been added. |
-| __quantum__rt__tuple_unreference | `void(%TupleHeader*)` | Indicates that an existing reference has been removed and potentially releases the tuple. |
+| __quantum__rt__tuple_create      | `%Tuple*(i64)`  | Allocates space for a tuple requiring the given number of bytes, sets the reference count to 1 and the user count to 0. |
+| __quantum__rt__tuple_add_user   | `void(%Tuple*)` | Increases the current user count by one. |
+| __quantum__rt__tuple_remove_user | `void(%Tuple*)` | Decreases the current user count by one, fails if the user count becomes negative. |
+| __quantum__rt__tuple_copy      | `%Tuple*(%Tuple*, i1)`  | Creates a shallow copy of the tuple if the user count is larger than 0 or the second argument is `true`. |
```
irinayat-MS, line 198: Perhaps, could explicitly say here that the calling code is responsible for proper reference counts updates on the contents of the tuple, because the runtime cannot do that?

irinayat-MS:
In the generated code from "" I'm also seeing array_get_size_1d instead of array_get_length. I'm fine with adding a specialized call to get length of 1d arrays, but shouldn't the naming be consistent with the corresponding method for multi-dim arrays?

Specifications/QIR/Data-Types.md
bettinaheim: Yes, a compiler may choose not to track alias counts and still need the option to have a copy be made.

irinayat-MS (dismissed):
lgtm (but it would be helpful to have more concrete PR title)

Specifications/QIR/Callables.md
```diff
@@ -351,13 +351,14 @@ callable values:
...
-| __quantum__rt__callable_reference | `void(%Callable*)`                      | Indicates that a new reference has been added. |
-| __quantum__rt__callable_unreference | `void(%Callable*)`                      | Indicates that an existing reference has been removed and potentially releases the callable value. |
+| __quantum__rt__callable_update_reference_count | `void(%Callable*, i64)`                      | Adds the given integer value to the reference count for the callable. Deallocates the callable if the reference count becomes 0. The behavior is undefined if the reference count becomes negative. |
+| __quantum__rt__callable_update_alias_count | `void(%Callable*, i64)`                      | Adds the given integer value to the alias count for the callable. Fails if the count becomes negative. |
+| __quantum__rt__callable_memory_management | `void(i32, %Callable*, i64)`                      | Invokes the function at the given index in the memory management table of the callable with the capture tuple and the given 64-bit integer. Does nothing if if the memory management table pointer or the function pointer at that index is null.  |
```
irinayat-MS, line 361: I'd put %Callable* argument first to be consistent with other methods in the table (and with the convention of passing 'this' pointer) but I don't feel strongly about it. Your call.

Specifications/QIR/Callables.md
```diff
@@ -318,6 +312,12 @@ language-specific compiler.
 The lambda body may need to include additional code that performs argument tuple
 construction before calling the underlying callable.
 
+#### Memory Management Table
+
+Any captured values need to remain alive as long as the callable value exists, and correspondingly may also need to be unreferenced when the type information for the captured values cannot be statically determined. Upon creation, a table with two function pointers is hence associated with a callable value.
```
irinayat-MS, line 317: Maybe, explain here that the memory management is recursive? (Otherwise, the callable could manage the capture tuple automatically and it's not clear why these additional facilities are needed.)

Specifications/QIR/Callables.md
```diff
@@ -318,6 +312,12 @@ language-specific compiler.
 The lambda body may need to include additional code that performs argument tuple
 construction before calling the underlying callable.
 
+#### Memory Management Table
+
+Any captured values need to remain alive as long as the callable value exists, and correspondingly may also need to be unreferenced when the type information for the captured values cannot be statically determined. Upon creation, a table with two function pointers is hence associated with a callable value. 
+
+Like the implementation table, the table is defined as global constant with a unique name. It contains two pointers of type `void(%Tuple*, i64)*`; the first one points to the function for modifying the reference counts of captured values, the second points to the one for modifying the access counts. They can be invoked using the runtime function `__quantum__rt__callable_memory_management`. If there are no captured values, a null pointer should be passed upon callable creation.
```
irinayat-MS, line 319: "access counts" -> "alias counts"

Specifications/QIR/Data-Types.md
```diff
@@ -96,8 +94,7 @@ simple types:
 | Function                          | Signature                | Description |
 |-----------------------------------|--------------------------|-------------|
 | __quantum__rt__result_equal       | `i1(%Result*, %Result*)` | Returns true if the two results are the same, and false if they are different. |
-| __quantum__rt__result_reference   | `void(%Result*)`         | Increments the reference count of a Result pointer. |
-| __quantum__rt__result_unreference | `void(%Result*)`         | Decrements the reference count of a Result pointer and releases the result if appropriate. |
+| __quantum__rt__result_update_reference_count   | `void(%Result*, i64)` | Adds the given integer value to the reference count for the result. Deallocates the result if the reference count becomes 0. The behavior is undefined if the reference count becomes negative. |
```
irinayat-MS, line 97: I've recently standardized the runtime to use i32 for ref/alias counts. Any reason we should be supporting i64?

Specifications/QIR/Quantum-Runtime.md
```diff
@@ -14,9 +14,9 @@ We define the following functions for managing qubits:
 | Function                            | Signature       | Description |
 |-------------------------------------|-----------------|-------------|
 | __quantum__rt__qubit_allocate       | `%Qubit*()`     | Allocates a single qubit. |
-| __quantum__rt__qubit_allocate_array | `%Array*(i64)`  | Allocates an array of qubits. |
+| __quantum__rt__qubit_allocate_array | `%Array*(i64)`  | Creates an array of the given size and populates it with newly allocated qubits. |
 | __quantum__rt__qubit_release        | `void(%Qubit*)` | Release a single qubit. |
-| __quantum__rt__qubit_release_array  | `void(%Array*)` | Release an array of qubits. |
+| __quantum__rt__qubit_release_array  | `void(%Array*)` | Release each qubit in the given array. The array itself is *not* unreferenced, and needs to be dereferenced separately for the memory to be freed. |
```
irinayat-MS, line 19: Maybe rename to something like `__quantum__rt__qubit_release_all`?

irinayat-MS (approved):

Specifications/QIR/Data-Types.md
```diff
@@ -2,15 +2,13 @@
```
kuzminrobin, line 2: Can you please also address the following?

1. I feel the discrepancy for the PaulyY and PaulyZ values between the fragment 
> 0 is PauliI, 1 is PauliX, _**3**_ is PauliY, and _**2**_ is PauliZ

and the fragment (sorry, the `@` causes some unexpected formatting)
> @PauliI = constant i2 0
@PauliX = constant i2 1
@PauliY = constant i2 _**-1**_
@PauliZ = constant i2 _**-2**_
2. I propose the following change.
```diff
2c2
< 0..2..4 = {0. 2. 4}
---
> 0..2..4 = {0, 2, 4}
```

kuzminrobin (commented):
FYI: https://github.com/microsoft/qsharp-language/pull/67/files#r566437736

Specifications/QIR/Data-Types.md
kuzminrobin: I understood the item 1. I propose the following clarification in [this file](https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Data-Types.md):
```diff
@@ -86,8 +86,8 @@ The following global constants are defined for use with the `%Result` and `%Paul
 
 @PauliI = constant i2 0
 @PauliX = constant i2 1
-@PauliY = constant i2 -1
-@PauliZ = constant i2 -2
+@PauliY = constant i2 -1    ; The value 3 (binary 11) is displayed as a 2-bit sigend value of -1 (binary 11).
+@PauliZ = constant i2 -2    ; The value 2 (binary 10) is displayed as a 2-bit sigend value of -2 (binary 10).
```

Specifications/QIR/Data-Types.md
bettinaheim: rephrased

Specifications/QIR/Data-Types.md
bettinaheim: changed to alias

Specifications/QIR/Data-Types.md
bettinaheim: I added a sentence that the reference count of tuple items remains unchanged, and also added a similar sentence to the other copy functions.

Specifications/QIR/Callables.md
bettinaheim: good catch

Specifications/QIR/Data-Types.md
bettinaheim: No, no reason besides that we are using i64 by default throughout QIR.

Specifications/QIR/Data-Types.md
bettinaheim: Thanks for catching the dot, and I added the comment.

Specifications/QIR/Data-Types.md
```diff
@@ -170,53 +165,45 @@ big integers.
...
+- The size of the tuple in bytes
+- The alias count indicating how many handles to the tuple exist in the source code
 
-The following utility functions are provided by the classical runtime to support
-tuples and user-defined types:
+The language specific compiler is responsible for injecting calls to increase and decrease the alias count as needed, as well as to accurately reflect when references to the LLVM structure representing a tuple are created and removed. 
+See the section [above](#reference-and-alias-counting) regarding the distinction between alias and reference counting. 
+
+In the case where the source language treats tuples as immutable values, the language-specific compiler is expected to request the necessary copies prior to modifying the tuple in place. 
+This is done by invoking the runtime function `__quantum__rt__tuple_copy` to create a byte-by-byte copy of a tuple. Unless the copying is forced via the second argument, the runtime may omit copying the value and instead simply return a pointer to the given argument if the alias count is 0 and it is hence save to modify the tuple in place.
```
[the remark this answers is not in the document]
bettinaheim: thanks

Specifications/QIR/Callables.md
bettinaheim: I expanded a bit more here.

Specifications/QIR/Callables.md
```diff
@@ -291,7 +287,7 @@ have access to the controlled count, and so can't unambiguously determine the
 expected argument tuple; specifically, it can't tell if an inner tuple is the
 result of an application of `Controlled` or just part of the base signature.
 
-#### Implementing Lambdas
+### Implementing Lambdas, Partial Application and Currying
```
alan-geller, line 290:
```suggestion
### Implementing Lambdas, Partial Application. and Currying
```

Specifications/QIR/Callables.md
```diff
@@ -318,6 +312,15 @@ language-specific compiler.
 The lambda body may need to include additional code that performs argument tuple
 construction before calling the underlying callable.
 
+#### Memory Management Table
+
+Since any captured values need to remain alive as long as the callable value exists, they also need to be unreferenced when the callable value is released. While sufficient type information for the captured values is known upon creation of the value, the information is no longer available at the time when it is released.
+Upon creation, a table with two function pointers for modifying reference and alias counts for captured values is hence associated with a callable value. 
+
+Like the implementation table, the table is defined as global constant with a unique name. It contains two pointers of type `void(%Tuple*, i64)*`; the first one points to the function for modifying the reference counts of captured values, the second points to the one for modifying the alias counts. Either of those pointers may be null, and if no values were captured, a null pointer should be passed instead of a table upon callable creation.
```
alan-geller, line 320:
```suggestion
Like the implementation table, the table is defined as global constant with a unique name. It contains two pointers of type `void(%Tuple*, i64)*`; the first one points to the function for modifying the reference counts of captured values, the second points to the one for modifying the alias counts. Either of those pointers may be null, and if no managed values were captured, a null pointer should be passed instead of a table upon callable creation.
```

I added "managed" because a null pointer should be also provided if all that's captured are some integers or other scalars.

Specifications/QIR/Callables.md
```diff
@@ -318,6 +312,15 @@ language-specific compiler.
...
+
+Like for tuple and array items, the responsibility of managing the reference and access count for captured values lays with the compiler. The two functions can be invoked using the runtime function `__quantum__rt__callable_memory_management`, see the description [below](https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Callables.md#runtime-functions).
```
alan-geller, line 322:
```suggestion
As for tuple and array items, the responsibility of managing the reference and alias count for captured values lays with the compiler. The two functions can be invoked using the runtime function `__quantum__rt__callable_memory_management`, see the description [below](https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Callables.md#runtime-functions). 
```

Specifications/QIR/Callables.md
```diff
@@ -351,13 +354,14 @@ callable values:
...
-| __quantum__rt__callable_reference | `void(%Callable*)`                      | Indicates that a new reference has been added. |
-| __quantum__rt__callable_unreference | `void(%Callable*)`                      | Indicates that an existing reference has been removed and potentially releases the callable value. |
+| __quantum__rt__callable_update_reference_count | `void(%Callable*, i64)`                      | Adds the given integer value to the reference count for the callable. Deallocates the callable if the reference count becomes 0. The behavior is undefined if the reference count becomes negative. |
+| __quantum__rt__callable_update_alias_count | `void(%Callable*, i64)`                      | Adds the given integer value to the alias count for the callable. Fails if the count becomes negative. |
+| __quantum__rt__callable_memory_management | `void(i32, %Callable*, i64)`                      | Invokes the function at the given index in the memory management table of the callable with the capture tuple and the given 64-bit integer. Does nothing if if the memory management table pointer or the function pointer at that index is null.  |
```
alan-geller, line 364: I would rename this `__quantum__rt__callable_manage_memory`, since generally all of our runtime function names are verbs or verb phrases. I might even split it into two separate functions, `__quantum__rt__callable_update_capture_reference` and `__quantum__rt__callable_update_capture_alias`.

Specifications/QIR/Callables.md
alan-geller, line 322: It might be worth giving an example where the ref count or alias count of the capture tuple elements could change without the ref count or access count of the callable changing. The simplest I can come up with is pretty trivial, in pseudo-Q#-ish:
`let f = lambda (x : Int[], y : Int) -> (x, y);`
Then every call to `f` has to increment the ref and access count of the capture tuple elements, specifically of the `x` array.

Specifications/QIR/Data-Types.md
```diff
@@ -1,42 +1,11 @@
 ## Data Type Representation
 
 We define LLVM representations for a variety of classical and quantum data types.
+QIR does not expect the runtime to provide garbage collection. Instead, it specifies a set of runtime functions that can be used by the language specific compiler to implement a reference counting scheme if needed, if the source language requires automatic memory management. See the section of [reference and alias counting](#reference-and-alias-counting) for more detail.
```
alan-geller, line 4: I might say "require" instead of "expect". I might also add a sentence saying that a runtime implementation that provides fully automated garbage collection is free to ignore (or strip out) the reference counting calls.

Specifications/QIR/Data-Types.md
```diff
@@ -170,53 +136,46 @@ big integers.
...
-```LLVM
-%TupleHeader = type { i32 }
-```
-
-For instance, a tuple containing two integers, `(Int, Int)`, would be represented in
-LLVM as `type {%TupleHeader, i64, i64}`.
+When [invoking callable values](https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Callables.md#invoking-a-callable-value) using the `__quantum__rt__callable_invoke` runtime function, 
+tuples are passed as a pointer to an opaque LLVM structure, `%Tuple`. The pointer is expected to point to the contained data such that it can be cast to the correct data structures by the
+receiving code.
+This permits to define runtime functions that are common for all tuples such as the ones listed below.
```
alan-geller, line 145:
```suggestion
This allows the definition of runtime functions that are common for all tuples such as the ones listed below.
```

alan-geller, line 145:
```suggestion
This permits the definition of runtime functions that are common for all tuples, such as the functions listed below.
```

Specifications/QIR/Data-Types.md
```diff
@@ -170,53 +136,46 @@ big integers.
...
+in LLVM depends on its usage.
+If used as a return type for a callable, it should be translated into an LLVM
+`void` function.
+If it is used as a value, for instance as an element of a tuple, it should be represented as a null tuple pointer.
 
 ### Arrays
 
-Array data is represented as a pointer to an opaque LLVM structure, `%Array`.
+Within QIR, arrays are represented and passed around as a pointer to an opaque LLVM structure, `%Array`. 
+How to represent array data, i.e. what that pointer points to, is at the discretion of the runtime. All array manipulations, including item access, hence need to be performed by invoking the corresponding runtime function(s).
```
alan-geller, line 178:
```suggestion
How array data is represented, i.e., what that pointer points to, is at the discretion of the runtime. All array manipulations, including item access, hence need to be performed by invoking the corresponding runtime function(s).
```

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
+| __quantum__rt__array_slice       | `%Array*(%Array*, i32, %Range)`      | Creates and returns an array that is a slice of an existing array. The `i32` indicates which dimension the slice is on, which must be smaller than the number of dimensions in the array. The `%Range` specifies the slice. |
 | __quantum__rt__array_project     | `%Array*(%Array*, i32, i64)`         | Creates and returns an array that is a projection of an existing array. The `i32` indicates which dimension the projection is on, and the `i64` specifies the specific index value to project. |
 
 There are special runtime functions defined for allocating or releasing an
 array of qubits.
 See [here](Quantum-Runtime.md#qubit-management-functions) for these functions.
 
+### Reference and Alias Counting
+
+QIR specifies a set of runtime functions for types that are represented as pointers that may be used by the language-specific compiler to expose them as immutable types in the language.
```
alan-geller, line 260: A minor nit is that these functions don't apply to qubits, which are represented by pointers. It might be worth pointing out that this section doesn't apply to qubits, just for clarity.

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
 array of qubits.
 See [here](Quantum-Runtime.md#qubit-management-functions) for these functions.
 
+### Reference and Alias Counting
+
+QIR specifies a set of runtime functions for types that are represented as pointers that may be used by the language-specific compiler to expose them as immutable types in the language. 
+
+To ensure that unnecessary copying of data can be avoided, QIR distinguishes two kinds of counts that can be tracked: reference counts and alias counts. 
+
+Reference counts track the number of handles that allow to access a certain value *in LLVM*. They hence determine when the value can be released by the runtime; values are allocated with a reference count of 1, and will be released when their reference count reaches 0.
```
alan-geller, line 264:
```suggestion
Reference counts track the number of handles that allow access to a certain value *in LLVM*. They hence determine when the value can be released by the runtime; values are allocated with a reference count of 1, and will be released when their reference count reaches 0. 
```

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
 
+### Reference and Alias Counting
+
+QIR specifies a set of runtime functions for types that are represented as pointers that may be used by the language-specific compiler to expose them as immutable types in the language. 
+
+To ensure that unnecessary copying of data can be avoided, QIR distinguishes two kinds of counts that can be tracked: reference counts and alias counts. 
+
+Reference counts track the number of handles that allow to access a certain value *in LLVM*. They hence determine when the value can be released by the runtime; values are allocated with a reference count of 1, and will be released when their reference count reaches 0. 
+
+Alias counts, on the other hand, track how may handles exist *in the source language*.
```
alan-geller, line 266:
```suggestion
Alias counts, on the other hand, track how many handles to a value exist *in the source language*. 
```

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
+### Reference and Alias Counting
+
+QIR specifies a set of runtime functions for types that are represented as pointers that may be used by the language-specific compiler to expose them as immutable types in the language. 
+
+To ensure that unnecessary copying of data can be avoided, QIR distinguishes two kinds of counts that can be tracked: reference counts and alias counts. 
+
+Reference counts track the number of handles that allow to access a certain value *in LLVM*. They hence determine when the value can be released by the runtime; values are allocated with a reference count of 1, and will be released when their reference count reaches 0. 
+
+Alias counts, on the other hand, track how may handles exist *in the source language*. 
+They determine when the runtime needs to copy data; when copy functions are invoked, the copy is executed only if the alias count is larger than 0, or the copy is explicitly forced. Alias counts are hence useful to optimize the handling of data types that are represented as pointers in QIR, but are value types, i.e. immutable, within the source language.
```
alan-geller, line 267:
```suggestion
They determine when the runtime needs to copy data; when copy functions are invoked, the copy is executed only if the alias count is larger than 0, or the copy is explicitly forced. Alias counts are useful for optimizing the handling of data types that are represented as pointers in QIR, but are value types, i.e. immutable, within the source language. 
```

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
+
+Reference counts track the number of handles that allow to access a certain value *in LLVM*. They hence determine when the value can be released by the runtime; values are allocated with a reference count of 1, and will be released when their reference count reaches 0. 
+
+Alias counts, on the other hand, track how may handles exist *in the source language*. 
+They determine when the runtime needs to copy data; when copy functions are invoked, the copy is executed only if the alias count is larger than 0, or the copy is explicitly forced. Alias counts are hence useful to optimize the handling of data types that are represented as pointers in QIR, but are value types, i.e. immutable, within the source language. 
+
+The compiler is responsible to track both counts as needed by injecting the corresponding calls to modify them. A call to modify such counts will only ever modify the count for the given instance itself and not for inner item such as e.g. tuple or array items, or captured values for callables; the compiler is responsible for injecting calls to update counts for inner items as needed.
```
alan-geller, line 269:
```suggestion
The compiler is responsible for generating code that tracks both counts correctly by injecting the corresponding calls to modify them. A call to modify such counts will only ever modify the count for the given instance itself and not for any inner items such as an element of a tuple or an array or a value captured by a callable; the compiler is responsible for injecting calls to update counts for inner items as needed.
```

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
+A targeted runtime is free to provide other mechanism for garbage collection and treat calls to modify reference counts as hints or as simple no-ops.
```
alan-geller, line 270:
```suggestion
A runtime implementation is free to provide another mechanism for garbage collection and to treat calls to modify reference counts as hints or as simple no-ops.
```

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
+
+- Runtime routines that create a new instance always initialize the instance
+  with a reference count of 1, and an alias count of 0.
+- For each pointer type, with the exception of `%Qubit*`, 
+  a runtime function ending in `_update_reference_count` exists that can be used to modify the reference count of an instance as needed. If the reference count reaches 0, the instance may be released. Decreasing the reference count below 0 or accessing a value after its reference count has reached 0 results in undefined behavior.
```
alan-geller, line 275: A philosophical note: I really dislike the notion of "undefined behavior" since it tends to result in the growth of unplanned spec items: a couple of influential implementations happen to pick the same behavior for some case that's undefined, and all of a sudden people are writing code that relies on that behavior.
I would much prefer to fail when something nonsensical occurs (negative reference count).
In the case of accessing after release, there's no way to force any particular behavior, so "undefined" is the best we can do.

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
+- For all data types that support a runtime function to create a shallow copy, 
+  a runtime function ending in `_update_alias_count` exists that can be used to modify the alias count of an instance as needed. These functions exist for `%Tuple*`, `%Array*`, and `%Callable*` types. The alias count can never be negative; decreasing the alias count below 0 results in a runtime exception.
```
alan-geller, line 277:
```suggestion
  a runtime function ending in `_update_alias_count` exists that can be used to modify the alias count of an instance as needed. These functions exist for `%Tuple*`, `%Array*`, and `%Callable*` types. The alias count can never be negative; decreasing the alias count below 0 results in program failure.
```
We don't have exceptions.

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
+- The function to modify reference and alias count accept a
```
alan-geller, line 278:
```suggestion
- The functions that modify reference and alias count should accept a 
```

Specifications/QIR/Data-Types.md
```diff
@@ -286,13 +245,38 @@ The following utility functions are provided if multidimensional array support i
...
+
+- Runtime routines that create a new instance always initialize the instance
+  with a reference count of 1, and an alias count of 0.
+- For each pointer type, with the exception of `%Qubit*`, 
+  a runtime function ending in `_update_reference_count` exists that can be used to modify the reference count of an instance as needed. If the reference count reaches 0, the instance may be released. Decreasing the reference count below 0 or accessing a value after its reference count has reached 0 results in undefined behavior.
+- For all data types that support a runtime function to create a shallow copy, 
+  a runtime function ending in `_update_alias_count` exists that can be used to modify the alias count of an instance as needed. These functions exist for `%Tuple*`, `%Array*`, and `%Callable*` types. The alias count can never be negative; decreasing the alias count below 0 results in a runtime exception.
+- The function to modify reference and alias count accept a 
+  null instance pointer and should simply ignore the call if the pointer is null.
```
alan-geller, line 279:
```suggestion
  null instance pointer and simply ignore the call if the pointer is null.
```

Specifications/QIR/Quantum-Runtime.md
```diff
@@ -14,9 +14,9 @@ We define the following functions for managing qubits:
 | Function                            | Signature       | Description |
 |-------------------------------------|-----------------|-------------|
 | __quantum__rt__qubit_allocate       | `%Qubit*()`     | Allocates a single qubit. |
-| __quantum__rt__qubit_allocate_array | `%Array*(i64)`  | Allocates an array of qubits. |
+| __quantum__rt__qubit_allocate_array | `%Array*(i64)`  | Creates an array of the given size and populates it with newly allocated qubits. |
```
alan-geller, line 17:
```suggestion
| __quantum__rt__qubit_allocate_array | `%Array*(i64)`  | Creates an array of the given size and populates it with newly-allocated qubits. |
```

alan-geller (approved):
Mostly grammatical and other small language tweaks, only one or two substantive comments.

Specifications/QIR/Data-Types.md
bettinaheim: I agree with that I would rather require a proper failure, but I think that is too much to ask, given that the runtime should be allowed to free and forget that memory.

Specifications/QIR/Data-Types.md
bettinaheim: Well, but the runtime has exceptions (usually), so I was thinking of requiring the runtime to throw. I am not sure what handling we would want to require if we didn't want to fail execution?

alan-geller: I would require failure. We require failure in some other situations, I believe.

Specifications/QIR/Quantum-Runtime.md
bettinaheim: I reconsidered what that function should do and correspondingly keep the name.

Specifications/QIR/Data-Types.md
bettinaheim: Changed to require, and the sentence you suggested is already there further below in the section.

Specifications/QIR/Data-Types.md
bettinaheim: I added a sentence.

Specifications/QIR/Callables.md
bettinaheim: Let's go with Alan's suggestion below and actually specify which function has to be at which index and give them distinct names.

Specifications/QIR/Data-Types.md
bettinaheim: I think you are right, i32 is enough and we are already using it in other places. I updated to i32.

Specifications/QIR/Callables.md
bettinaheim: I gave each one separate names.

Specifications/QIR/Callables.md
bettinaheim: I assume you meant to have x be captured in this example. This example wouldn't actually require going via the capture update function every time. the only time you really need to update the capture counts is when you update the callable counts. However, neither the compiler not the runtime has enough type information to do that unless we save those functions with the callable, and it just seems more consistent with how other counts work to have count changes always apply only to the exact value given as argument, and not anything "inside" that value.

PR #68, opened 2020-12-23
Assign IDs for QEPs: Q# Enhancement Proposals

samarsha:
I also made some tweaks to the proposal template:

* Move the proposal title into a level 1 header and increase the level of every existing header by 1
* Use a standard date format, YYYY-MM-DD
* Use standard frontmatter syntax instead of an HTML comment


bettinaheim (approved):
Thanks for setting this up!

Issue #70, opened 2021-01-05
Implementation QEP 1


bettinaheim:
**Implementation**

This issue tracks the implementation of [QEP 1: Implicitly-Scoped Qubit Allocation](https://github.com/microsoft/qsharp-language/blob/main/Approved/1-implicitly-scoped-qubit-allocation.md#qep-1-implicitly-scoped-qubit-allocation).

*Work items*

* [x] The feature branch for the work can be found [here](https://github.com/microsoft/qsharp-compiler/tree/feature/inline-qubit-scope)
* [x] The work items are tracked in microsoft/qsharp-compiler#795
* [x] Merge to main (microsoft/qsharp-compiler#815)
* [x] Update the Q# language specification

PR #71, opened 2021-01-12
Update Q# specification for QEP 1

### LINKED ISSUE #47: Specification should say that allocated qubits can be released early if possible
The page on [Quantum Memory Management](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/QuantumMemoryManagement.md#quantum-memory-management) says this about `using` statements:

> The `using`-statement allocates the qubits from the quantum processor's free qubit heap, and returns them to the heap after the statement terminates.

This should be updated to say that the qubits could be released *before* the end of the statement if the qubits are not used after a certain point in the block. The same could apply to the `borrowing` statement as well:

> The borrowed qubits are in an unknown state and go out of scope at the end of the statement block.

samarsha:
See #70. Also closes #47.


Specifications/Language/2_Statements/QuantumMemoryManagement.md
```diff
@@ -16,52 +16,67 @@ At the end of the block, the bound variables go out of scope and are no longer d
...
-using ((aux, register) = (Qubit(), Qubit[5])) {
+use (aux, register) = (Qubit(), Qubit[5]) {
     // ...
 }
 ```
 
 The qubits are guaranteed to be in a |0⟩ state upon allocation. They are released at the end of the scope and are required to either be in a |0⟩ state upon release, or to have been measured right beforehand. This requirement is not compiler-enforced, since this would require a symbolic evaluation that quickly gets prohibitively expensive. When executing on simulators, the requirement can be runtime enforced. On quantum processors, the requirement cannot be runtime enforced; an unmeasured qubit may be reset to |0⟩ via unitary transformation. Failing to do so will result in incorrect behavior. 
 
-The `using`-statement allocates the qubits from the quantum processor's free qubit heap, and returns them to the heap after the statement terminates.
+The `use`-statement allocates the qubits from the quantum processor's free qubit heap, and returns them to the heap no latter than the end of the scope in which the qubits are bound.
```
bettinaheim, line 48: Is "latter" opposed to "later" here correct?

bettinaheim (dismissed):

Specifications/Language/2_Statements/QuantumMemoryManagement.md
samarsha: No, that's a typo, thanks!

bettinaheim (approved):

PR #72, opened 2021-01-12
Update Q# specification for optional parentheses in callable types and curried calls

samarsha:
This PR updates the specification and grammar to match these changes to the compiler:

* microsoft/qsharp-compiler#710
* microsoft/qsharp-compiler#789


bettinaheim (approved):
I am glad we (you) got rid of the need for these parenthesis! :)

PR #73, opened 2021-01-14
Fix links replacement script to handle /tree/ links

tcNickolas:
Currently the script handles only links of the form `https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/`, but the documentation also contains links with `/tree/` instead of `/blob/` which are left unchanged. 

This means that on the docs website part of the links point to other pages on the same website and part of the links point to this GitHub repo (for example, the link "example program" on [this page](https://docs.microsoft.com/en-us/quantum/user-guide/language/programstructure/callabledeclarations) or type links [here](https://docs.microsoft.com/en-us/quantum/user-guide/language/expressions/valueliterals)). 

It makes sense to unify the behavior for links that are included in the uid mapping.

### DIFF
--- a/Specifications/sync-script/replace_links.py
+++ b/Specifications/sync-script/replace_links.py
@@ -39,10 +39,13 @@ def links_to_xrefs(source_dir, map_link_to_uid):
             for line in text:
                 for url in map_link_to_uid:
                     uid = map_link_to_uid[url]
-                    full_url = 'https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/'+url
+                    # Handle both /blob/ and /tree/ links
+                    full_blob_url = 'https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/'+url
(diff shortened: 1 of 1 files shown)


tcNickolas:
@bettinaheim Do you know who's a good person to review this? Would be nice to fix the way the links are converted before the next release.

cgranade (approved):

geduardo (approved):
Good catch! Thanks! I didn't think about this when creating the script. Although the "ideal" solution would be to always use "blob" and not "tree" since bolbs are the objects. But those two lines should work and solve the problem :).

tcNickolas:
It's always better to do a one-line technical solution than to require different people working on the docs to follow the requirement :-)

geduardo:
This wasn't enough to fix it since the tree links weren't in the uid map. I opened a PR to add them: https://github.com/MicrosoftDocs/quantum-docs-pr/pull/1158

Hopefully that would work.

PR #83, opened 2021-03-03
QIR specification update 1

bettinaheim:


Specifications/QIR/Callables.md
```diff
@@ -291,7 +287,7 @@ have access to the controlled count, and so can't unambiguously determine the
 expected argument tuple; specifically, it can't tell if an inner tuple is the
 result of an application of `Controlled` or just part of the base signature.
 
-#### Implementing Lambdas
+## Implementing Lambdas, Partial Application. and Currying
```
bettinaheim, line 290:
```suggestion
## Implementing Lambdas, Partial Application, and Currying
```

Specifications/QIR/Quantum-Runtime.md
```diff
@@ -1,47 +1,26 @@
...
 
 | Function                            | Signature       | Description |
 |-------------------------------------|-----------------|-------------|
 | __quantum__rt__qubit_allocate       | `%Qubit*()`     | Allocates a single qubit. |
-| __quantum__rt__qubit_allocate_array | `%Array*(i64)`  | Allocates an array of qubits. |
-| __quantum__rt__qubit_release        | `void(%Qubit*)` | Release a single qubit. |
-| __quantum__rt__qubit_release_array  | `void(%Array*)` | Release an array of qubits. |
+| __quantum__rt__qubit_allocate_array | `%Array*(i64)`  | Creates an array of the given size and populates it with newly-allocated qubits. |
+| __quantum__rt__qubit_release        | `void(%Qubit*)` | Releases a single qubit. |
+| __quantum__rt__qubit_release_array  | `void(%Array*)` | Releases and array of qubits; each qubit in the array is released, and the array itself is unreferenced. |
```
alan-geller, line 19:
```suggestion
| __quantum__rt__qubit_release_array  | `void(%Array*)` | Releases an array of qubits; each qubit in the array is released, and the array itself is unreferenced. |
```

Specifications/QIR/Quantum-Runtime.md
```diff
@@ -1,47 +1,26 @@
...
-accessed while they are borrowed.
-The code that borrows the qubits guarantees that the state of the qubits when
-returned is identical, including entanglement, to their state when borrowed.
-It is always acceptable to satisfy `borrow` by allocating new qubits.
-
-*__Discussion__*
->It will likely be useful to provide usage hints to `alloc` and `borrow`.
->Since we don't know yet what form these hints may take, we leave defining them
->for future work.
+Any measurements or resets applied upon release are at the discretion of the target as well; the qubits that are to be released hence need to be unentangled from qubits that remain live prior to invoking the release function.
```
alan-geller, line 23: I might reword the second phrase a bit here. There's no QIR requirement that the qubits be unentangled; we won't check that, and we won't generate an error if they are. Rather, behavior of the quantum algorithm might be unpredictable if qubits are released that are still entangled with active program qubits.

Specifications/QIR/Data-Types.md
```diff
@@ -1,44 +1,31 @@
-## Data Type Representation
+# Data Type Representation
 
-We define LLVM representations for a variety of classical and quantum data types.
+QIR defines an LLVM representations for a variety of classical and quantum data types that may be used as part of a compiled quantum program. For more information about classical memory management including reference and alias counting, see [here](Classical-Runtime.md#memory-management).
```
alan-geller, line 3:
```suggestion
QIR defines LLVM representations for a variety of classical and quantum data types that may be used as part of a compiled quantum program. For more information about classical memory management including reference and alias counting, see [here](Classical-Runtime.md#memory-management).
```

Specifications/QIR/Data-Types.md
```diff
@@ -1,44 +1,31 @@
...
-This allows each target to provide a structure definition appropriate for that
-target.
+## Opaque Types
 
-### Reference Counting
+Representing certain types as pointers to opaque LLVM structure types allows each target to provide a structure definition appropriate for that target.
 
-In QIR, all types represented as pointers, other than qubits, are reference-counted.
-All types follow the same pattern:
+In addition to arrays and callable values, the type `%Qubit*` used to represent qubits, and the type `%Result*` representing measurement results, are both opaque within QIR.
```
alan-geller, line 9:
```suggestion
In addition to arrays and callable values, the types `%Qubit*` that represents qubits and `%Result*` that represents measurement results are both opaque within QIR. 
```

alan-geller (approved):
A few minor grammatical things.

Specifications/QIR/Data-Types.md
```diff
@@ -271,28 +226,28 @@ arrays:
...
 | __quantum__rt__array_get_element_ptr_1d | `i8*(%Array*, i64)`           | Returns a pointer to the element of the array at the zero-based index given by the `i64`. |
-| __quantum__rt__array_slice       | `%Array*(%Array*, i32, %Range)`      | Creates and returns an array that is a slice of an existing array. The `i32` indicates which dimension the slice is on, which must be 0 for a 1-dimensional array. The `%Range` specifies the slice. |
-| __quantum__rt__array_reference   | `void(%Array*)`                      | Indicates that a new reference has been added. |
-| __quantum__rt__array_unreference | `void(%Array*)`                      | Indicates that an existing reference has been removed and potentially releases the array. |
+| __quantum__rt__array_update_reference_count   | `void(%Array*, i32)` | Adds the given integer value to the reference count for the array. Deallocates the array if the reference count becomes 0. The behavior is undefined if the reference count becomes negative. |
+| __quantum__rt__array_update_alias_count | `void(%Array*, i32)` | Adds the given integer value to the alias count for the array. Fails if either count becomes negative. |
```
kuzminrobin, line 235: (For negative ref count the _behavior is undefined_. But for negative alias count - _fail_)
Just out of curiosity...
My understanding is that if the count becomes negative then it means an internal bug for both cases - ref count and alias count. Then why the behavior is different (if the count becomes negative)? Why not both fail or both undefined behavior?

Specifications/QIR/Code-Generation.md
```diff
@@ -7,8 +7,7 @@ from a simple rewrite of basic intrinsics to target machine code:
   garbage collection, and thus carefully tracks stack versus heap allocation
   and reference counting for heap-allocated structures. A runtime that provides
   full garbage collection may wish to remove the reference count field from several
-  intermediate representation structures and elide calls to `__quantum__rt__free`
-  and the various `unreference` functions.
+  intermediate representation structures and elide calls the various `unreference` functions.
 - Many types are defined as pointers to opaque structures. The code generator
```
kuzminrobin, line 11: (Missing word?) 
I assume you meant "and elide calls **to** the various `unreference` functions"

Specifications/QIR/Classical-Runtime.md
```diff
@@ -1,29 +1,41 @@
...
+
+- Runtime routines that create a new instance always initialize the instance
+  with a reference count of 1, and an alias count of 0.
+- For each pointer type, with the exception of `%Qubit*`, 
+  a runtime function ending in `_update_reference_count` exists that can be used to modify the reference count of an instance as needed. If the reference count reaches 0, the instance may be released. Decreasing the reference count below 0 or accessing a value after its reference count has reached 0 results in undefined behavior.
+- For all data types that support a runtime function to create a shallow copy, 
+  a runtime function ending in `_update_alias_count` exists that can be used to modify the alias count of an instance as needed. These functions exist for `%Tuple*`, `%Array*`, and `%Callable*` types. The alias count can never be negative; decreasing the alias count below 0 results in a runtime failure.
+- The functions that modify reference and alias count should accept a 
+  null instance pointer and simply ignore the call if the pointer is null.
```
kuzminrobin, line 38: Just out of curiosity.
May I know the cases when this is needed? (accessing the null instance)

Specifications/QIR/Callables.md
```diff
@@ -344,20 +347,22 @@ The LLVM representation treats these generated callables as the actual
...
-| __quantum__rt__callable_reference | `void(%Callable*)`                      | Indicates that a new reference has been added. |
-| __quantum__rt__callable_unreference | `void(%Callable*)`                      | Indicates that an existing reference has been removed and potentially releases the callable value. |
+| __quantum__rt__callable_update_reference_count | `void(%Callable*, i32)`                      | Adds the given integer value to the reference count for the callable. Deallocates the callable if the reference count becomes 0. The behavior is undefined if the reference count becomes negative. |
+| __quantum__rt__callable_update_alias_count | `void(%Callable*, i32)`                      | Adds the given integer value to the alias count for the callable. Fails if the count becomes negative. |
+| __quantum__rt__capture_update_reference_count | `void(%Callable*, i32)`                      | Invokes the function at index 0 in the memory management table of the callable with the capture tuple and the given 64-bit integer. Does nothing if if the memory management table pointer or the function pointer at that index is null.  |
```
kuzminrobin, line 364: Does the fragment "the given 64-bit integer" relate to the second parameter of type `i32`?
I feel that some words are missing for the fragment "and the given 64-bit integer". Something like " ~~and~~ **to add** the given 64-bit integer **to the reference count of ..**". 

Typo? 
"if if"

Applicable to `__quantum__rt__capture_update_alias_count ` below also.

kuzminrobin (approved):
The changes in this PR definitely improve the documentation.
Approving with some minor notes and questions.

Specifications/QIR/Data-Types.md
bettinaheim: I'd prefer to have both fail properly - undefined behavior is always harder to debug - but given that the change decreasing the ref count below zero may be applied to something that is already deallocated, it doesn't quite come for free to do so. Being required to properly fail I believe may have performance implications and would require additional work in the runtime; if you see any clever solutions for avoiding that, I am all ears. :)

Specifications/QIR/Classical-Runtime.md
bettinaheim: The section on the Memory Management Table asks to set the pointers to null if no values are captured that would require any count changes. At the same time, the compiler can't always know whether that is the case at the time when it needs to invoke a function to change the count, since that depends on the callable value and is thus runtime information in general.

Specifications/QIR/Data-Types.md
kuzminrobin: Thank you, Bettina. I really wasn't taking into account that the attempt to decrease the ref count down to a negative value can be accessing the _already deallocated_ instance. And such an access (even the _read_ of a ref count of the deallocated instance) already causes the _undefined behavior_ (e.g. the compiler can fully optimize out any such access if the compiler knows for sure that the instance is already deallocated).

Specifications/QIR/Callables.md
```diff
@@ -143,28 +140,27 @@ For the example above, the following would be generated:
...
+    void (%Tuple*, %Tuple*, %Tuple*)*
         @Some__Namespace__Symbol__adj,
-    void (%TupleHeader*, %TupleHeader*, %TupleHeader*)* 
+    void (%Tuple*, %Tuple*, %Tuple*)* 
         @Some__Namespace__Symbol__ctl,
-    void (%TupleHeader*, %TupleHeader*, %TupleHeader*)* 
+    void (%Tuple*, %Tuple*, %Tuple*)* 
         @Some__Namespace__Symbol__ctladj
   ]
 ```
```
kuzminrobin, line 154: @Some__Namespace__Symbol = constant 
  [void (%Tuple*, %Tuple*, %Tuple*)\*]
  [
    void (%Tuple*, %Tuple*, %Tuple*)*
        @Some__Namespace__Symbol__body **__wrapper**,
    void (%Tuple*, %Tuple*, %Tuple*)*
        @Some__Namespace__Symbol__adj **__wrapper**,
    void (%Tuple*, %Tuple*, %Tuple*)* 
        @Some__Namespace__Symbol__ctl **__wrapper**,
    void (%Tuple*, %Tuple*, %Tuple*)* 
        @Some__Namespace__Symbol__ctladj **__wrapper**
  ]

Specifications/QIR/Callables.md
```diff
@@ -1,9 +1,9 @@
-## Callables
+# Callables
 
 We use to term _callable_ to mean a subroutine in the source language.
 Different source languages use different names for this concept.
 
-### Basics
+## Basics
```
kuzminrobin, line 7: Would be nice to see either a definition or a link explaining what exactly _functor_ means in this context. The meaning that I intuitively feel (but not necessarily correctly understand) here, differs from the meaning that I remember from the C++ books.

Specifications/QIR/Callables.md
```diff
@@ -256,12 +252,12 @@ containing the array of control qubits as the first element and a tuple of
...
+`{ %Tuple, %Array*, { %Tuple, i64, %Qubit* }* }`, and the twice-`Controlled`
 version expects
-`{ %TupleHeader, %Array*, { %TupleHeader, %Array*, { %TupleHeader, i64, %Qubit* }* }* }`.
+`{ %Tuple, %Array*, { %Tuple, %Array*, { %Tuple, i64, %Qubit* }* }* }`.
 The "ctl" implementation function always expects
-`{ %TupleHeader, %Array*, { %TupleHeader, i64, %Qubit* }* }`.
+`{ %Tuple, %Array*, { %Tuple, i64, %Qubit* }* }`.
 Thus, if the controlled count is greater than 1,
 `__quantum__rt__callable_invoke` needs to disassemble the argument tuple,
 concatenate the control qubit arrays, and form the expected argument tuple.
```
kuzminrobin, line 263: **Important.**
_In this particular paragraph_ replacing the `%TupleHeader` with `%Tuple` is a wrong/misleading thing to do.
For clarity I would recommend to remove any mentioning of `%Tuple..` inside of `{ }` in this paragraph, i.e. to replace 
`{ %Tuple, i64, %Qubit* }` with `{ i64, %Qubit* }` and so on.

If anybody feels uncomfortable regarding such a change, then to leave `{ %TupleHeader, i64, %Qubit* }`. Because this reflects the real internals of the tuple. In this case the `%TupleHeader` will need to be defined somewhere earlier.

kuzminrobin: This is also applicable for 2 more paragraphs below.

Specifications/QIR/Callables.md
```diff
@@ -291,7 +287,7 @@ have access to the controlled count, and so can't unambiguously determine the
 expected argument tuple; specifically, it can't tell if an inner tuple is the
 result of an application of `Controlled` or just part of the base signature.
 
-#### Implementing Lambdas
+## Implementing Lambdas, Partial Application, and Currying
 
 The language-specific compiler should generate a new top-level callable of the
 appropriate type with implementation provided by the anonymous body of the lambda;
```
kuzminrobin, line 293: (These lines are not a part of a change in this PR, that is why feel free to ignore. Just in case I would like to make sure that you have this in mind)

Since this is a part of the public documentation, the external readers may have the same questions that we have. If we don't resolve those questions beforehand, then multiple users can ask us the same questions.

> should generate a new **top-level** callable

I'm not sure I correctly get the point of "top-level" fragment here. Is it a "global scope" callable? (as opposed to the _local scope_ callable)

Specifications/QIR/Quantum-Runtime.md
bettinaheim: Rephrased.

Specifications/QIR/Data-Types.md
```diff
@@ -133,7 +105,7 @@ allocated by the caller because the length of the string depends on the actual v
 | __quantum__rt__range_to_string   | `%String*(%Range)`   | Returns a string representation of the range. |
 | __quantum__rt__bigint_to_string  | `%String*(%BigInt*)` | Returns a string representation of the big integer. |
 
-### Big Integers
+## Big Integers
```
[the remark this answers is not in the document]
bettinaheim: The spec is not primarily meant to reflect what we implement, but is meant as a shared convention for different implementers. What is and isn't implemented should hence be documented as part of our compiler and runtime rather than on the spec here.

Specifications/QIR/Data-Types.md
```diff
@@ -168,55 +139,48 @@ big integers.
...
+- The alias count indicating how many handles to the tuple exist in the source code
 
-In some situations, a pointer to a tuple of unknown or variable type may be required.
-To satisfy LLVM's strong typing in such a situation, tuple pointers are passed as
-pointers to the initial tuple header field; that is, as a `%TupleHeader*`.
-These pointers should be cast to pointers to the correct data structures by the
-receiving code.
-For instance, this convention is used for callable wrapper functions; see
-[below](#callable-values-and-wrapper-functions).
+The language specific compiler is responsible for injecting calls to increase and decrease the alias count as needed, as well as to accurately reflect when references to the LLVM structure representing a tuple are created and removed.
```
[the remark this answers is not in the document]
bettinaheim: Yes, that is correct. However, to avoid the impression that QIR is in any way specific to Q# - which it is explicitly not meant to be - I'd rather avoid mention Q# over any other language that may choose to use this as their IR.

Specifications/QIR/Callables.md
bettinaheim: Yes, edited to say "callable at the global scope".

Specifications/QIR/Callables.md
bettinaheim: Thanks for catching this! I indeed overlooked this when I did the edits to update the tuple representation.

Specifications/QIR/Callables.md
bettinaheim: Indeed functors here are a somewhat quantum specific concept. I am not sure there are good links to just add here, and I'll leave it for a separate PR to potentially elaborate and explain more.

Specifications/QIR/Callables.md
```diff
@@ -143,28 +140,27 @@ For the example above, the following would be generated:
 
 ```LLVM
 @Some__Namespace__Symbol = constant
```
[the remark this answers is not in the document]
bettinaheim: Edited as suggested.

alan-geller (approved):

PR #85, opened 2021-03-08
Fix snippet in copy and update expressions article

geduardo:
Fixes https://github.com/MicrosoftDocs/quantum-docs/issues/65


bettinaheim (approved):
Thanks for the catch.

PR #86, opened 2021-03-10
Rename "expression statement" to "call statement"

tcNickolas:
The term "expression statement" is used only once in the language spec and points to a page detailing "call statements", so it looks like a leftover from a previous naming convention. C# and Java define "expression statements" but those are broader than Q# "call statements" and include assignments and new object creation, so using these terms interchangeably might be confusing.

(If my reading is incorrect and these are intended to be used as synonyms, feel free to close the PR)

### DIFF
--- a/Specifications/Language/2_Statements/README.md
+++ b/Specifications/Language/2_Statements/README.md
@@ -4,8 +4,8 @@ Q# distinguishes between statements and expressions. Q# programs consist of a mi
 
 The following statements are currently available in Q#:
 
-* **Expression Statement**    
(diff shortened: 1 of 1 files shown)


samarsha:
Q# compiler internals also refer to it as an expression statement, since it's a statement that contains an expression and nothing else. There's the additional restriction that the expression must be a call expression, but that restriction could be lifted in the future (especially if Q# goes expression-based). Maybe for now it's less confusing to use the more specific term though. @bettinaheim?

bettinaheim:
> Q# compiler internals also refer to it as an expression statement, since it's a statement that contains an expression and nothing else. There's the additional restriction that the expression must be a call expression, but that restriction could be lifted in the future (especially if Q# goes expression-based). Maybe for now it's less confusing to use the more specific term though. @bettinaheim?

Yes, I agree that calling it a call statement until/unless we remove the restriction in the future is less confusing.

bettinaheim (approved):

PR #93, opened 2021-04-29
Fix typo in Comments page

tcNickolas:
A couple more suggestions I think could improve the page:
 * It would be good to have a link to https://docs.microsoft.com/en-us/azure/quantum/contributing-docs, since it offers a more detailed example with multiple sections, formulas etc.
 * Does it make sense to mention that we can use LaTeX in doc comments but it won't be rendered in IntelliSense, so Unicode can be better for the portions that show up in quick help?

Let me know if I should add these to the PR as well.


bettinaheim:
Sorry, just saw this. Brad's copy edits make this obsolete, since he fixed the typo on there as well. Thanks!

tcNickolas:
Great! What about the suggestions in the PR description? It doesn't look like those are in main, so just double-checking whether you saw them.

PR #105, opened 2021-06-11
Added the preliminary list of reserved quantum operations

kuzminrobin:
Continuation of the branch [here](https://github.com/microsoft/qsharp-language/pull/101)


Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,41 @@
...
+| T | [`__quantum__qis__t__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L224) | The [T operation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.t) | [Single-qubit operations](https://docs.microsoft.com/en-us/azure/quantum/concepts-the-qubit#single-qubit-operations), [Rotations](https://docs.microsoft.com/en-us/azure/quantum/user-guide/libraries/standard/prelude#rotations). ![T = \begin{bmatrix} 1 & 0 \\ 0 & e^{i\pi/4} \\ \end{bmatrix} = \begin{bmatrix} 1 & 0 \\ 0 & \frac{1}{\sqrt{2}}+ \frac{1}{\sqrt{2}}i \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+T+%3D+%5Cbegin%7Bbmatrix%7D+1+%26+0+%5C%5C+0+%26+e%5E%7Bi%5Cpi%2F4%7D+%5C%5C+%5Cend%7Bbmatrix%7D+%3D+%0A%5Cbegin%7Bbmatrix%7D+1+%26+0+%5C%5C+0+%26+%5Cfrac%7B1%7D%7B%5Csqrt%7B2%7D%7D%2B+%5Cfrac%7B1%7D%7B%5Csqrt%7B2%7D%7Di+%5C%5C+%5Cend%7Bbmatrix%7D) |
+
+| Operation Name | QIR Implementation | Description | Notes |
```
alan-geller, line 22: I would merge the two tables.

Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,41 @@
...
+In this table, operation names should be considered case-insensitive.
+That is, if a target implements `RX` instead of `Rx`, its effect still must match that
+defined in the table.
+
+The matrices in this tables below take some time to load.
+Please be patient!
+
+| Operation Name | QIR Implementation | Description | Notes |
+|----------------|-----------------|-------------|-------|
+| H | [`__quantum__qis__h__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L127) | The [Hadamard operation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.h) | [Single-qubit operations](https://docs.microsoft.com/en-us/azure/quantum/concepts-the-qubit#single-qubit-operations), [Other Single-Qubit Cliffords](https://docs.microsoft.com/en-us/azure/quantum/user-guide/libraries/standard/prelude#other-single-qubit-cliffords) |
```
alan-geller, line 19: I think rather than the link, I would put the explicit matrix here. Partly that's because the QIR specification shouldn't depend on the Q# documentation, and partly that's because the matrix is the precise definition of the gate.

Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,41 @@
...
+| S | [`__quantum__qis__s__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L198) | The [S operation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.s) (Phase Gate)  | [Single-qubit operations](https://docs.microsoft.com/en-us/azure/quantum/concepts-the-qubit#single-qubit-operations), [Other Single-Qubit Cliffords](https://docs.microsoft.com/en-us/azure/quantum/user-guide/libraries/standard/prelude#other-single-qubit-cliffords). ![S = T^2](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+S+%3D+T%5E2) |
```
alan-geller, line 25: The algebraic relationships between the operations aren't really necessary. The operation matrices specify the operations completely.

Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,41 @@
...
+| Rx | [`__quantum__qis__r__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L148), see [R](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.r) | The [Rx operation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.rx) | [Single-qubit operations](https://docs.microsoft.com/en-us/azure/quantum/concepts-the-qubit#single-qubit-operations), [Rotations](https://docs.microsoft.com/en-us/azure/quantum/user-guide/libraries/standard/prelude#rotations) |
```
alan-geller, line 30: These three should have different QIR implementation names: `__quantum__qis__rx__body`, `__quantum__qis__ry__body`, and `__quantum__qis__rz__body`. The descriptions should be stand-alone; for instance, Rx is the X rotation gate.

alan-geller (commented):
I would include the explicit matrix for each operation, and avoid any links to Q# documentation. QIR is intended to stand alone and be useful separate from Q#, so it's better if the QIR spec doesn't reference Q# documentation.

kuzminrobin:
> I would include the explicit matrix for each operation, and avoid any links to Q# documentation. QIR is intended to stand alone and be useful separate from Q#, so it's better if the QIR spec doesn't reference Q# documentation.

I would like to cover 2 aspects here.
1. My impression was that we need to unblock one of the hardware vendors ASAP. So the preliminary list does not need to be a final version. It needs to be a brief list we can quickly refer the vendor to. So in this preliminary list I tried to avoid text and matrix duplication as much as possible.
@bettinaheim, please provide your opinion about that. Shall I start moving towards the final version of the list _right now_?
2. The cited fragment gives me somewhat new understanding. Let me summarize.
Until recently my understanding was that QIR (Quantum Intermediate Representation) is generated to work in pair with the QIR RT (QIR Runtime), which is a part of the Q# compiler. And I was considering all 3 tightly tied together.
Recently I hear more and more about using the QIR RT with other languages (which until now seemed to me more theoretical, for a remote prospective). And your comment, @alan-geller, sort of pushed me over some threshold after which I started thinking deeper about the languages other than Q#.
Now my understanding is that the QIR and QIR RT are considered as a product potentially fully separate from Q# compiler. By example of Figure 11.3 in [Intro to LLVM](http://www.aosabook.org/en/llvm.html), multiple frontends for multiple languages are considered/expected to generate the QIR (a subset of the LLVM IR), and in combination with the QIR RT the generated QIR will be used by multiple backends to generate the byte code for multiple quantum architectures.
@alan-geller, @bettinaheim, please correct me if that understanding is not quite right.

Specifications/QIR/Reserved-Operations.md
kuzminrobin: I can do that, it is easy. 

I just wanted to emphasize that the H and T operations cannot be implemented with other operations. But the others can be implemented based on H and T (but I'm not sure if the rotations can be implemented with H and T). 

When recently I needed to study the fundamentals of the quantum computing, it was curious for me to realize that a number of operations can be implemented based on a very small set of the other basic operations. And no any piece of information was providing an explicit picture/table of such a layered structure of commands. I just tried to make a step in that direction, to give a hint to the new hardware vendors, that their implementation can be simplified if they implement the quantum operation in a certain order.

For the same purposes I added the formulas showing the dependencies between the operations, like `Z = T^4 = S^2 = HXH`.

alan-geller: Even though that is true mathematically -- and is true for some other sets of operations as well, not just H and T -- it's not how most actual hardware works. The simplest, easiest to implement operations vary wildly depending on the type of hardware, and so hardware vendors are constrained by that rather than by the algebraic structure of the operations.

I would have one table, sorted alphabetically.

cgranade: > I just wanted to emphasize that the H and T operations cannot be implemented with other operations. But the others can be implemented based on H and T (but I'm not sure if the rotations can be implemented with H and T).

The `H`,  `S`, `CNOT`, and `T` operations are universal for compiling arbitrary precision implementations of arbitrary operations. That may not be the particular decomposition used by a particular implementer, however — some targets may validly treat `H` as derived from other operations unique to that target, or may choose another non-Clifford operation such as `V` instead of `T` to build universal gate sets.

alan-geller:
> > I would include the explicit matrix for each operation, and avoid any links to Q# documentation. QIR is intended to stand alone and be useful separate from Q#, so it's better if the QIR spec doesn't reference Q# documentation.
> 
> I would like to cover 2 aspects here.
> 
> 1. My impression was that we need to unblock one of the hardware vendors ASAP. So the preliminary list does not need to be a final version. It needs to be a brief list we can quickly refer the vendor to. So in this preliminary list I tried to avoid text and matrix duplication as much as possible.
>    @bettinaheim, please provide your opinion about that. Shall I start moving towards the final version of the list _right now_?
> 2. The cited fragment gives me somewhat new understanding. Let me summarize.
>    Until recently my understanding was that QIR (Quantum Intermediate Representation) is generated to work in pair with the QIR RT (QIR Runtime), which is a part of the Q# compiler. And I was considering all 3 tightly tied together.
>    Recently I hear more and more about using the QIR RT with other languages (which until now seemed to me more theoretical, for a remote prospective). And your comment, @alan-geller, sort of pushed me over some threshold after which I started thinking deeper about the languages other than Q#.
>    Now my understanding is that the QIR and QIR RT are considered as a product potentially fully separate from Q# compiler (despite the fact that the QIR documentation is in the **qsharp**-language repo, and the QIR RT implementation is in the **qsharp**-runtime repo). By example of Figure 11.3 in [Intro to LLVM](http://www.aosabook.org/en/llvm.html), multiple frontends for multiple languages are considered/expected to generate the QIR (a subset of the LLVM IR), and in combination with the QIR RT the generated QIR will be used by multiple backends to generate the byte code for multiple quantum architectures.
>    @alan-geller, @bettinaheim, please correct me if that understanding is not quite right.

On your second point: yes, your understanding is correct. While the QIR spec is hosted in the Q# language repo for now, our intention is that QIR should be independent of Q# and usable with any quantum language. We are actually working with at least one partner who is using QIR as a back-end for other quantum languages already.

Specifications/QIR/Reserved-Operations.md
kuzminrobin: I assume that with "algebraic relationships" you mean the fragments like this "S = T^2".
I explained that in the [earlier comment](https://github.com/microsoft/qsharp-language/pull/105#discussion_r651227848).

Please feel free to let me know if you would prefer me to remove those fragments.

Specifications/QIR/Reserved-Operations.md
kuzminrobin: I assume that with "These three" you mean "Z, X, Y" (the selection shows four: "Z, X, Y, Rx").

Then, I'm not strong in quantum computing, but my understanding is that [X](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.x) and [Rx](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.rx) are different operations (maybe X is a particular case of Rx, but I'm not sure). That is why it would be wrong thing if X operation was implemented under the name `__quantum__qis__rx__body`.

Please provide more details.

alan-geller: Sorry, the selection wasn't quite right -- the three are Rx, Ry, and Rz, which should be implemented under 3 different names.

Specifications/QIR/Reserved-Operations.md
kuzminrobin: Thank you for explaining, Alan and Chris. That is very good to know :-). 
I'll merge the tables and sort alphabetically.

Specifications/QIR/Reserved-Operations.md
alan-geller: Personally I would leave them out. It is unlikely to be useful to a hardware vendor (since the way they implement these operations is unlikely to depend on the algebraic relations) and doesn't add precision. The matrices provide exactness, everything else is extraneous.

kuzminrobin:
I guess I resolved all the feedback. Please review again.

Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,38 @@
...
+| Y | [`__quantum__qis__y__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L263) | The Y operation | ![\begin{bmatrix} 0 & -i \\ i & 0 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+0+%26+-i+%5C%5C+i+%26+0+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
+| Z | [`__quantum__qis__z__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L276) | The Z operation (Parity Gate) | ![\begin{bmatrix} 1 & 0 \\ 0 & -1 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+1+%26+0+%5C%5C+0+%26+-1+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
```
alan-geller, line 33:
```suggestion
| Z | [`__quantum__qis__z__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L276) | The Pauli Z operation (Parity Gate) | ![\begin{bmatrix} 1 & 0 \\ 0 & -1 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+1+%26+0+%5C%5C+0+%26+-1+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
```

Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,38 @@
...
+| X | [`__quantum__qis__x__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L250) | The X operation (NOT Gate) | ![\begin{bmatrix} 0 & 1 \\ 1 & 0 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+0+%26+1+%5C%5C+1+%26+0+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
+| Y | [`__quantum__qis__y__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L263) | The Y operation | ![\begin{bmatrix} 0 & -i \\ i & 0 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+0+%26+-i+%5C%5C+i+%26+0+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
```
alan-geller, line 32:
```suggestion
| Y | [`__quantum__qis__y__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L263) | The Pauli Y operation | ![\begin{bmatrix} 0 & -i \\ i & 0 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+0+%26+-i+%5C%5C+i+%26+0+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
```

Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,38 @@
...
+| Tdg | [`__quantum__qis__t__adj`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L230) | The adjoint T operation or T Dagger (![T^\dagger](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+T%5E%5Cdagger)) | ![\begin{bmatrix} 1 & 0 \\ 0 & e^{-i\pi/4} \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+1+%26+0+%5C%5C+0+%26+e%5E%7B-i%5Cpi%2F4%7D+%5C%5C+%5Cend%7Bbmatrix%7D) | See ![T^\dagger](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+T%5E%5Cdagger) [here](https://en.wikipedia.org/wiki/Clifford_gates#Building_a_universal_set_of_quantum_gates) |
+| X | [`__quantum__qis__x__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L250) | The X operation (NOT Gate) | ![\begin{bmatrix} 0 & 1 \\ 1 & 0 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+0+%26+1+%5C%5C+1+%26+0+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
```
alan-geller, line 31:
```suggestion
| X | [`__quantum__qis__x__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L250) | The Pauli X operation (NOT Gate) | ![\begin{bmatrix} 0 & 1 \\ 1 & 0 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cbegin%7Bbmatrix%7D+0+%26+1+%5C%5C+1+%26+0+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
```

Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,38 @@
...
+| H | [`__quantum__qis__h__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L127) | The Hadamard operation | ![\frac{1}{\sqrt{2}}\begin{bmatrix} 1 & 1 \\ 1 & -1 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cfrac%7B1%7D%7B%5Csqrt%7B2%7D%7D%5Cbegin%7Bbmatrix%7D+1+%26+1+%5C%5C+1+%26+-1+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
+| Measure | [`__quantum__qis__measure__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L140) | The Measure operation |  |
+| Reset |  | The Reset operation |  |
```
alan-geller, line 23:
```suggestion
| Reset | `__quantum__qir__reset_body`  | The Reset operation, which prepares the qubit in the Z=0 state  |  |
```

Specifications/QIR/Reserved-Operations.md
```diff
@@ -0,0 +1,38 @@
...
+| H | [`__quantum__qis__h__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L127) | The Hadamard operation | ![\frac{1}{\sqrt{2}}\begin{bmatrix} 1 & 1 \\ 1 & -1 \\ \end{bmatrix}](https://render.githubusercontent.com/render/math?math=%5Cdisplaystyle+%5Cfrac%7B1%7D%7B%5Csqrt%7B2%7D%7D%5Cbegin%7Bbmatrix%7D+1+%26+1+%5C%5C+1+%26+-1+%5C%5C+%5Cend%7Bbmatrix%7D) |  |
+| Measure | [`__quantum__qis__measure__body`](https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Qir/Runtime/lib/QSharpCore/qsharp-core-qis.ll#L140) | The Measure operation |  |
```
alan-geller, line 22: I would replace this with Mz (measure in the Pauli Z basis). The general, multiqubit Measure operation in Q# is actually pretty unusual, while Mz is standard.

alan-geller (commented):
A couple of last suggestions... Looks good, thanks!!

Specifications/QIR/Reserved-Operations.md
kuzminrobin: Replacing.

Specifications/QIR/Reserved-Operations.md
kuzminrobin: Done.

Specifications/QIR/Reserved-Operations.md
kuzminrobin: Done

Specifications/QIR/Reserved-Operations.md
kuzminrobin: Done

alan-geller:
> > > I would include the explicit matrix for each operation, and avoid any links to Q# documentation. QIR is intended to stand alone and be useful separate from Q#, so it's better if the QIR spec doesn't reference Q# documentation.
> > 
> > 
> > I would like to cover 2 aspects here.
> > 
> > 1. My impression was that we need to unblock one of the hardware vendors ASAP. So the preliminary list does not need to be a final version. It needs to be a brief list we can quickly refer the vendor to. So in this preliminary list I tried to avoid text and matrix duplication as much as possible.
> >    @bettinaheim, please provide your opinion about that. Shall I start moving towards the final version of the list _right now_?
> > 2. The cited fragment gives me somewhat new understanding. Let me summarize.
> >    Until recently my understanding was that QIR (Quantum Intermediate Representation) is generated to work in pair with the QIR RT (QIR Runtime), which is a part of the Q# compiler. And I was considering all 3 tightly tied together.
> >    Recently I hear more and more about using the QIR RT with other languages (which until now seemed to me more theoretical, for a remote prospective). And your comment, @alan-geller, sort of pushed me over some threshold after which I started thinking deeper about the languages other than Q#.
> >    Now my understanding is that the QIR and QIR RT are considered as a product potentially fully separate from Q# compiler (despite the fact that the QIR documentation is in the **qsharp**-language repo, and the QIR RT implementation is in the **qsharp**-runtime repo). By example of Figure 11.3 in [Intro to LLVM](http://www.aosabook.org/en/llvm.html), multiple frontends for multiple languages are considered/expected to generate the QIR (a subset of the LLVM IR), and in combination with the QIR RT the generated QIR will be used by multiple backends to generate the byte code for multiple quantum architectures.
> >    @alan-geller, @bettinaheim, please correct me if that understanding is not quite right.
> 
> On your second point: yes, your understanding is correct. While the QIR spec is hosted in the Q# language repo for now, our intention is that QIR should be independent of Q# and usable with any quantum language. We are actually working with at least one partner who is using QIR as a back-end for other quantum languages already.

@bettinaheim , do you want to express an opinion on linking to Q# documentation in this page, at least for now? I think that's the last question before merging...

Specifications/QIR/Reserved-Operations.md
kuzminrobin: Replacing the implementation column with the following:
> Library implementation: [Reset](https://github.com/microsoft/qsharp-runtime/blob/cff5574d69e7fd9bbdb496b1c4a85fd45723425e/src/Simulation/TargetDefinitions/Decompositions/Reset.qs#L19). Hardware override: [Reset](https://github.com/microsoft/qsharp-runtime/blob/cff5574d69e7fd9bbdb496b1c4a85fd45723425e/src/Simulation/TargetDefinitions/Intrinsic/Reset.qs#L13).

bettinaheim:
@kuzminrobin "Now my understanding is that the QIR and QIR RT are considered as a product potentially fully separate from Q# compiler (despite the fact that the QIR documentation is in the qsharp-language repo, and the QIR RT implementation is in the qsharp-runtime repo)" They are indeed fully separate and independent entities. The repo setup as is is not great and we are currently exploring options for creating a fully separate repo for work around QIR. 
Hence, please do not refer to or mention Q# in any way as part of the QIR specs. While I appreciate the effort to be mindful of redundancies also in docs, the direction should go the other way around; QIR defines what meaning certain functions have if they are available, and what sets of instruction sets are commonly used. It should fully define that without linking to any external sources (including qsharp* repos or even the Q# specs here). To avoid redundancy, the Q# doc comment may choose to link to the definitions in the QIR specs for the corresponding operations.

swernli:
I think we need to be more careful about how this asserts the way controls are used, namely it calls out a correlation between `CNOT` and `__quantum__qis__x__ctl` that I don’t think holds, because the latter supports arbitrary controls and not just one. We need to be careful not to let Q# patterns or abstractions like this overly influence how we define QIR instructions or it may hamper usability. Specifically here, I think we should avoid asserting anything about controlled gates that take an array of controls, since that assumes the runtime for the target supports arrays.

kuzminrobin:
Agreed. I will make changes once I get back to this task.

PR #123, opened 2021-12-16
Update ANTLR lexer and grammar

yjt98765:
This pull request updates the ANTLR files for describing the language grammar. It mainly contains the recent [modification](https://github.com/microsoft/qsharp-compiler/pull/1285) in the [compiler repository](https://github.com/microsoft/qsharp-compiler). Some earlier modification to `QSharpParser.g4` is also added to keep the consistency between these two repositories.


bettinaheim:
@samarsha Could you take a quick look at this?

Specifications/Language/5_Grammar/QSharpParser.g4
```diff
@@ -217,8 +218,8 @@ expression
     | expression '&&&' expression
     | expression '^^^' expression
     | expression '|||' expression
-    | expression 'and' expression
-    | expression 'or' expression
+    | expression ('&&' | 'and') expression
+    | expression ('||' | 'or') expression
```
samarsha, line 222: Because `&&` and `||` are new tokens, they need to be added to QSharpLexer.g4 also.

samarsha (changes requested):
Hi @yjt98765, thanks for the PR.

While the grammar on qsharp-compiler targets C#, the grammar here targets Java. This means that with your changes the grammar now has a mix of C# and Java predicates and can't be compiled with either language target.

I think it would make sense to make all of the predicates C# for consistency with qsharp-compiler, but the other option is to convert the new predicates to Java.

yjt98765:
@samarsha Sorry, I did not notice the grammar targets Java. I have updated the lines you mentioned above and generated Java files using ANTLR 4.9.2. VS Code does not report errors regarding the generated files. As a comparison, the original predicate `_localctx.terminal.Text` did cause an error. Although I did not execute the parser for actually parsing a Q# file, I guess the previous evidence is strong enough.

Specifications/Language/5_Grammar/QSharpParser.g4
```diff
@@ -227,6 +228,8 @@ expression
     | expression 'w/' expression '<-' expression
     ;
 
+sizeKey : Identifier {_localctx.getText() == "size"}?;
```
bkiers, line 231: Given the target language is Java, `{_localctx.getText() == "size"}?` should be `{_localctx.getText().equals("size")}?`

yjt98765: Thank you!

samarsha (approved):
Looks good now, thanks!

PR #140, opened 2022-04-15
Add type inference page for Q# documentation

samarsha:
This also contains advice on dealing with ambiguous types, so that the Q# compiler can link to it in its ambiguous type error message (microsoft/qsharp-compiler#1095).


alan-geller (approved):
Looks good to me.

PR #154, opened 2022-08-03
Proposal for the Dyn Type

spignorel:
This proposal introduces a Dyn type to reflect values that are only known at runtime.

### DIFF
--- a/Approved/dyn-type.md
+++ b/Approved/dyn-type.md
@@ -1,0 +1,196 @@
+A Dyn Type For Runtime Values
+=====
+
+## Proposal
+
+This proposal introduces a `dyn` type to reflect values that are only known at runtime.
+
+## Description
+
+The `dyn` type doesn't require any user annotation as it is inferred by the compiler. The `dyn` type is used with global callables. `dyn` values come from intrinsic operations. Functions can't introduce `dyn` values, but can pass them. The Result type is always a `dyn`. A `dyn` Result can be converted to a `dyn` Bool.  The `dyn` static analysis replaces mutable variable and result analysis.
+
+## Use Cases
+
+Below use cases explain how a developer interacts with the `dyn` type and describe the functionalities of the new type.
+
+### Allow to pass the values created at runtime to other callables
+
+```qsharp
+operation Reset(q : Qubit, m : Result) : Unit {
+    if m == One {
+        X(q);
+    }
+}
+
+operation MeasureAndReset(q : Qubit) : Unit {
+    let res = M(q); // create a dynamic value (res)
+    Reset(q, res); // pass the dynamic value to another operation
+}
+```
+
+Without the proposed modification, the implementation of `Reset` would need to be manually inlined into `MeasureAndReset`.
+
+### Allow callables to return the values created at runtime
+
+```qsharp
+operation MeasureAsInt(q : Qubit) : Int {
+    return M(q) == One ? 1 | 0;
+}
+
+operation CountOnes(qs : Qubit[]) : Int {
+    mutable count = 0;
+    for q in qs {
+        set count += MeasureAsInt(q);
+    }
+    return count;
+}
+```
+
+Without the proposed modification, dynamic values can only be returned by the entry point, but not by other callables.
+
+Without introducing dynamic types, callables defined in libraries could only be used for computing compile time known values, but not for manipulating runtime values; Ultimately, application code then would need to be mostly contained in the entry point callable and code cannot be structured into several subroutines.
+
+### Dynamic values require runtime support
+
+The example generates an error when calling `Foo` with a dynamic value as argument `i` unless the target supports arrays.
+
+The developer can create values at runtime by measuring a qubit and then pass these values to other callables that process it further and return it.
+
+> Define the array index as a callable input parameter
+
+```qsharp
+operation Foo(b : Bool, i : Int) : Int {
+    return [1, 2, 3][i]; // This needs runtime support if i is dynamic
+}
+
+operation Bar() : Int {
+// get a dynamic value (usually by measuring a qubit)
+    use q = Qubit();
+    H(q);
+    let val = M(q) == Zero ? 0 : 1; // val is of type dyn Int
+
+    let res = Foo(true, val); // requires that the runtime supports arrays
+    return res; // res is of type dyn Int
+}
+```
+
+> Define the lifting operator for the array index
+
+```qsharp
+lift_i Foo : (Bool, dyn Int) => dyn Int
+```
+
+> The array index is lifted into dyn. The result value of the operation is also a dyn.
+
+
+### Lift a boolean into dyn
+
+> Define the boolean as a callable input parameter
+
+```qsharp
+operation Foo(b : Bool, i : Int) : Int {
+    use q = Qubit();
+    if b { X(q); }
+    return [1, 2, 3][i];
+}
+```
+
+> Define the lifting operator for the boolean
+
+```qsharp
+lift_b Foo : (dyn Bool, Int) => Int
+```
+
+> The boolean is lifted into `dyn`. The result value of the operation is not `dyn`. The result is opaque as `b` is used to branch for a gate.
+
+### Cannot lift anonymous parameters into dyn
+
+> Define anonymous parameters as a callable input parameter
+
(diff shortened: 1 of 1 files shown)


Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+
+The example generates an error when calling `Foo` with a dynamic value as argument `i` unless the target supports arrays.
+
+The developer can create values at runtime by measuring a qubit and then pass these values to other callables that process it further and return it.
+
+> Define the array index as a callable input parameter
+
+```qsharp
+operation Foo(b : Bool, i : Int) : Int {
+    return [1, 2, 3][i]; // This needs runtime support if i is dynamic
```
cgranade, line 63: What error would be returned to developers in this case? Is that included in the current spec for capability levels, or is there a new kind of capability that would be required?

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+
+Without the proposed modification, dynamic values can only be returned by the entry point, but not by other callables.
+
+Without introducing dynamic types, callables defined in libraries could only be used for computing compile time known values, but not for manipulating runtime values; Ultimately, application code then would need to be mostly contained in the entry point callable and code cannot be structured into several subroutines.
+
+### Dynamic values require runtime support
+
+The example generates an error when calling `Foo` with a dynamic value as argument `i` unless the target supports arrays.
+
+The developer can create values at runtime by measuring a qubit and then pass these values to other callables that process it further and return it.
```
cgranade, line 57: Are there any other ways to create values of type `dyn ★`? If so, may be good to list them here; if not, may be good to make explicit that this is exhaustive.

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+
+    let res = Foo(true, val); // requires that the runtime supports arrays
+    return res; // res is of type dyn Int
+}
+```
+
+> Define the lifting operator for the array index
+
+```qsharp
+lift_i Foo : (Bool, dyn Int) => dyn Int
```
cgranade, line 80: Is this literal Q# code? If so, that would seem to imply that `dyn` is part of the syntax?

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+}
+```
+
+> Define the lifting operator for the boolean
+
+```qsharp
+lift_b Foo : (dyn Bool, Int) => Int
+```
+
+> The boolean is lifted into `dyn`. The result value of the operation is not `dyn`. The result is opaque as `b` is used to branch for a gate.
```
cgranade, line 104: This makes it sound like the `dyn`-ness of an operation cannot be guaranteed from the signature alone, to @samarsha's point. Is that intended?

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Cannot lift anonymous parameters into dyn
+
+> Define anonymous parameters as a callable input parameter
+
+```qsharp
+operation Foo(Bool, Int) : Int {
+    use q = Qubit();
+    if b { X(q); }
```
cgranade, line 113: Where is `b` defined?

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+```qsharp
+If x : T, then inject x : dyn T
+```
+
+> The compiler automatically injects compile-time values into dyn
+
+## Implementation
+
+The implementation will be informed by below design goals:
+> Goal 1: Keep compilation and optimization logic within llvm as much as possible, rather than replicating it in the Q# compiler
```
cgranade, line 138:
```suggestion
> Goal 1: Keep compilation and optimization logic within LLVM as much as possible, rather than replicating it in the Q# compiler.
```

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+
+> Goal 2: Q# functions serve the same purpose as const functions in other languages, i.e. we basically want to guarantee that a function can be evaluated when and as soon as all arguments are known. This in particular means that unless an argument is `dyn`, we expect that the function call will be fully evaluated/evaluatable at compile time.
```
cgranade, line 140: Does that imply that a function can not return a `dyn` unless at least one input was `dyn`?

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Analyzer pass on each callable parameter
+
+The analyzers are run for one callable at a time (same as it is the case today). They will be run once in (more or less) the same way as they are today, and produce the same attribute they are today - i.e. the first pass of the analyzers analyzes the requirements of the callable under the assumption that all arguments are non-dyn values.
+
+Additionally, we will run the analyzers once for each parameter of the callable. This pass will produce an additional attribute per parameter that indicates the required runtime capability if a specific parameter is `dyn` instead of non-`dyn`. We hence end up with n+1 attribute on the callable, where n is the number of parameters. For each call to the callable, the required capability for that call is given by the sum of the required capabilities for the callable itself plus those for of the `dyn` arguments.
```
cgranade, line 155: What will these attributes be called? If they exist on the final AST output by the compiler, then they will need to be public parts of the Core library API and would then need API review as well.

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+
+The n+1 attributes need to be manually declared for intrinsic callables (same as it is the case today, just more attributes).
+
+## Approximations
+
+For the purpose of type inference (i.e. how `dyn` is propagated), we make the below approximations.
+
+### Operations always return a dyn
+
+The return type of an operation is always `dyn`. While it is conceivable to write an operation that has a non-`dyn` return value, it is also always possible - and arguably best practice - to split out any computation of a non-`dyn` return value into a function instead.
```
cgranade, line 165: Could we raise a warning or a lint message when an operation returns `dyn` only because of this rule? How will we document this approximation?

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+## Limitations
+
+The required target capability can be computed for each callable parameter separately, and the overall capability requirements for the callable is a sum (not the product!) of the capability requirement for each parameter + the capability requirement for the callable itself (i.e. the capability requirement due to runtime values the callable itself introduces).
+
+This in particular means that we do not intend to distinguish, now or in the future, if e.g. a targeted backend supports comparing a dynamic value against a constant, but not comparing two dynamic values with each other.
+
+Concretely, we do not intend to distinguish the ability to execute the following two pieces of code:
+
+```qsharp
+if (M(q) == Zero) {
```
cgranade, line 184:
```suggestion
if M(q) == Zero {
```

Approved/dyn-type.md
```diff
@@ -0,0 +1,196 @@
...
+    X(q);
+}
+```
+
+compared to
+
+```qsharp
+if (M(q1) != M(q2)) {
```
cgranade, line 192:
```suggestion
if M(q1) != M(q2) {
```

cgranade (commented):
Thanks for pinging me on this, @spignorel! I think my biggest overall question would be how we would document, explain, and expose this to users; a lot of the approximations, for instance, seem very reasonable but may take users by surprise unless we can offer good feedback as early as possible.

samarsha:
Although dyn as a type was originally my idea, I think my opinion now is that dyn should not be a type, but rather an internal compiler property for expressions. We don't plan to expose dyn to users (and potentially it may never need to be exposed, given our inference rules), so I think it would be better to avoid cementing it in Q#'s type system.

For the compiler's static analysis, I think we only need:
- A property on expressions that the compiler can track (there's already a mechanism for tracking "local quantum dependencies," which is like a weaker form of dyn - we can replace that with dyn).
- Attributes on the callable indicating the capabilities required when each parameter is dyn.

So I'm not sure if this should be a QEP at all, since we didn't write a QEP for the earlier static analysis of hardware capabilities before, but left it as an internal compiler implementation detail.

billti:
Are we in agreement on this now? @spignorel do you need to update the proposal based on the latest from @samarsha above?

spignorel:
Per @samarsha there are questions as to which repo this should sit under. @billti stay tuned.

PR #170, opened 2023-12-13
Updating Language Guide

ScottCarda-MS:
Updates the language guide to reflect changes to the Q# language made recently.


Specifications/Language/2_Statements/README.md
```diff
@@ -1,42 +1,20 @@
...
+* **Expression statement**    
+    Contains a Q# expression to be run, such as a call to an operation. If the last statement in a block is an expression statement, it may have its trailing semicolon omitted to give the block the evaluated value of the contained expression.
 
 * **Variable declaration**    
     Defines one or more local variables that are valid for the remainder of the current scope, and binds them to the specified values. Variables can be permanently bound or declared to be reassignable later on. See [Variable declarations and reassignments](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/2_Statements/VariableDeclarationsAndReassignments.md#variable-declarations-and-reassignments) for more details.
 
 * **Variable reassignment**
```
swernli, line 13: Technically, variable reassignments are an expression, not a statement.

Specifications/Language/3_Expressions/CallExpressions.md
```diff
@@ -0,0 +1,24 @@
+# Call expressions
+
+Call expressions are an important part of any programming language. Operation and function calls, much like [partial applications](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/3_Expressions/Closures.md#partial-application), can be used as an expression anywhere as long as the returned value is of a suitable type.
+
+The usefulness of calling functions in this form primarily lies in debugging, and such operation calls are one of the most common constructs in any Q# program. At the same time, operations can only be called from within other operations and not from within functions. For more information, see also [Qubits](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/4_TypeSystem/QuantumDataTypes.md#qubits).
```
swernli, line 5: "The usefulness of calling functions" sounds weird when the second half of the sentence says "operation calls" since we generally either draw a distinction between functions and operations or use a more generic term. Maybe something like "The usefulness of call expressions" just to stay consistent with the title?

swernli: Oh, I see the context now, the sentence realy is about operations. Maybe the first part should say "operation" instead of "function" or "call expressions using operations" or something similar.

Specifications/Language/3_Expressions/CallExpressions.md
```diff
@@ -0,0 +1,24 @@
...
+
+With callables being first-class values, call expressions are a generic way of supporting patterns that aren't common enough to merit their own dedicated language construct, or dedicated syntax has not (yet) been introduced for other reasons. Some examples of library methods that serve that purpose are `ApplyIf`, which invokes an operation conditional on a classical bit being set; `ApplyToEach`, which applies a given operation to each element in an array; and `ApplyWithInputTransformation`, as shown in the following sample.
```
swernli, line 7: I dont' think we have `ApplyIf` anymore, so its mention here should be removed.

Specifications/Language/3_Expressions/ComparativeExpressions.md
```diff
@@ -2,31 +2,26 @@
...
+Equality comparison of arrays, and tuples are supported by comparisons of their items, and are only supported if all of their nested types support equality comparison.
 
-## Quantitative comparison
+Equality comparison of ranges are supported, and two ranges are considered equal if xthey produce the same sequence of integers. For example, the following two ranges
```
swernli, line 14: Not sure if it's worth calling out, but open-ended ranges do not support equality, since they don't have a known, concrete integer sequence (it depends on how they are used for slicing).

Specifications/Language/3_Expressions/ConditionalBranching.md
```diff
@@ -35,9 +32,8 @@ You can also express simple branching in the form of a [conditional expression](
...
+The tight integration between control-flow constructs and quantum computations poses a challenge for current quantum hardware. Certain quantum processors do not support branching based on measurement outcomes. As such, comparison for values of type `Result` will always result in a compilation error for Q# programs that are targeted to run on such hardware.
 
+Other quantum processors support specific kinds of branching based on measurement outcomes. The more general `if` expressions supported in Q# are compiled into suitable instructions that can be run on such processors. The imposed restrictions are that values of type `Result` may only be compared as part of the condition within `if` expressions in operations. Furthermore, the conditionally run blocks cannot contain any `return` expressions or update mutable variables that are declared outside that block.
```
swernli, line 37: The language around not supporting return or mutable updates in blocks that are conditioned on measurement comparison is old language that is not true once the adaptive profile comes into play. I'm not sure if that is worth addressing now though.

Specifications/Language/2_Statements/README.md
ScottCarda-MS: Oh I thought they were both statements. That's unfortunate because the docs talk about them together, and it is going to be difficult to go through and disentangle them. For now, maybe I can just drop this paragraph?

Specifications/Language/3_Expressions/ConditionalLoops.md
```diff
@@ -36,21 +34,21 @@ until condition;
...
+The `repeat` loop runs a block of statements before evaluating a condition. If the condition evaluates to true, the loop exits. If the condition evaluates to false, an additional block of statements defined as part of an optional `fixup` block, if present, is run prior to entering the next loop iteration.
 
-The compiler treats all parts of the `repeat` statement (both blocks and the condition) as a single scope for each repetition; symbols that are defined within the `repeat` block are visible both to the condition and within the `fixup` block. As for other loops, symbols go out of scope after each iteration, such that symbols defined in the `fixup` block are not visible in the `repeat` block.
+The compiler treats all parts of the `repeat` loop (both blocks and the condition) as a single scope for each repetition; symbols that are defined within the `repeat` block are visible both to the condition and within the `fixup` block. As for other loops, symbols go out of scope after each iteration, such that symbols defined in the `fixup` block are not visible in the `repeat` block.
```
swernli, line 39: This one actually isn't true anymore... it was a hard feature to support and was unfortunately bugprone, so now each block in the repeat-expression is treated as it's own scope, similar to if-expressions (see [this code sample](https://microsoft.github.io/qsharp/?code=H4sIAAAAAAAAE12OvWrDQBCE%2B3uKATV2EyVtQiBFgiuDlaQzLtbS2l64P%2B72iI3RuwedgxGZ8pudmW3bFg2%2ByEXLpm1brFhV%2FBFZKSkPE6u8wTvnPklUCb6S75NkSAbBiRdHFl2DmMIxkYOeSNGTx55RMg%2FQcKvET5I60DXow8APxpPjHKlnrC9dIa%2FFbf5arsYAwNuH13TZBPG6WFYSIieaPsGaxC%2BWeMYn52J1u8O1XkxKHJl0Bia5orS3jDNe8fhyt0YUr2JxIJv5Tg9yLvFfQWat4adZeDapJXlsdzdzNKP5BSRayv1jAQAA)). The expected workaround is to declare any variables that need to be visible across the scopes outside of the repeat-expression.

Specifications/Language/2_Statements/README.md
swernli: Yeah, I think leaving them together makes sense, since splitting up declaring mutable variables and updating them makes the docs harder to navigate. I think dropping this paragraph is fine.

Specifications/Language/3_Expressions/CallExpressions.md
ScottCarda-MS: Unfortunately there are a lot of examples and references to library callables that we don't have anymore. The purpose of this PR is going to be specifically around the language syntax changes, to try to scope this PR.

Specifications/Language/3_Expressions/ConditionalLoops.md
```diff
@@ -36,21 +34,21 @@ until condition;
...
 
-Loops that break based on a condition are challenging to process on quantum hardware if the condition depends on measurement outcomes since the length of the instruction sequence to run is not known ahead of time. 
+Loops that break based on a condition are challenging to process on quantum hardware if the condition depends on measurement outcomes since the length of the instruction sequence to run is not known ahead of time.
 
 Despite their common presence in particular classes of quantum algorithms, current hardware does not yet provide native support for these kinds of control flow constructs. Running on quantum hardware can potentially be supported in the future by imposing a maximum number of iterations, or as additional hardware support becomes available.
 
 ## While loop
 
-A more familiar-looking statement for classical computations is the `while` loop. It is supported only within functions. 
+A more familiar-looking loop for classical computations is the `while` loop. It is supported only within functions.
```
swernli, line 49: This is also no longer true, while loops can now appear in operations. This makes explaining things easier, as operations are a strict superset of functions, and the compiler checks only need to verify that functions don't use any features they aren't allowed to use.

Specifications/Language/3_Expressions/ConditionalLoops.md
```diff
@@ -64,6 +62,6 @@ while index < Length(arr) && item < 0 {
 The statement block is run as long as the condition evaluates to `true`.
 
 > [!NOTE]
-> Due to the challenge they pose for execution, we discourage the use of loops that break based on a condition and hence do not support `while` loops within operations. The use of `while` loops within operations may be considered in the future, with the restriction that the condition cannot depend on the outcome of a quantum measurement. 
+> Due to the challenge they pose for execution, we discourage the use of loops that break based on a condition and hence do not support `while` loops within operations. The use of `while` loops within operations may be considered in the future, with the restriction that the condition cannot depend on the outcome of a quantum measurement.
```
swernli, line 65: Ah, here's the note about it :). Since while loops are now allowed in operations, this note can just be removed.

Specifications/Language/3_Expressions/ConditionalBranching.md
ScottCarda-MS: I'd rather not touch on that now, and we can re-evaluate the language once the adaptive profile is ready.

Specifications/Language/3_Expressions/CopyAndUpdateExpressions.md
```diff
@@ -12,63 +12,61 @@ They are constructed using a ternary operator `w/` `<-`; the syntax `w/` should
...
-- `arr w/ 0 <- 10` is the array `[10,1,2,3]`.
-- `arr w/ 2 <- 10` is the array `[0,1,10,3]`.
-- `arr w/ 0..2..3 <- [10,12]` is the array `[10,1,12,3]`.
+- `arr w/ 0 <- 10` is the array `[10, 1, 2, 3]`.
+- `arr w/ 2 <- 10` is the array `[0, 1, 10, 3]`.
+- `arr w/ 0..2..3 <- [10, 12]` is the array `[10, 1, 12, 3]`.
 
-Copy-and-update expressions allow the efficient creation of new arrays based on existing ones. 
-The implementation for copy-and-update expressions avoids copying the entire array 
+Copy-and-update expressions allow the efficient creation of new arrays based on existing ones.
```
swernli, line 59: I'm not sure I'd agree with the word "efficient" here, but that's not really in scope right now ;)

Specifications/Language/2_Statements/README.md
ScottCarda-MS: I've dropped this paragraph here to minimize confusion, but I think a larger rewrite is going to be necessary to get this correct.

Specifications/Language/3_Expressions/CallExpressions.md
ScottCarda-MS: I've changed "function" to "operation" per your suggestions.

Specifications/Language/3_Expressions/ComparativeExpressions.md
ScottCarda-MS: Added a sentence to this effect.

Specifications/Language/3_Expressions/ConditionalLoops.md
ScottCarda-MS: Okay, I think the idea that each of the blocks create their own scope is the expected behavior of blocks in general, so we should be fine to just drop the paragraph rather than try to fix it.

Specifications/Language/3_Expressions/ConditionalLoops.md
ScottCarda-MS: Is it just that functions don't allow repeat-until loops?

swernli: Yup. Functions can't use the following (as seen in [the pass that checks for these errors](https://github.com/microsoft/qsharp/blob/main/compiler/qsc_passes/src/callable_limits.rs)):
 - conjugate expressions
 - functor declarations
 - specialization declarations
 - calls to operations
 - qubit allocation statements
 - repeat-until loop expressions

Operations no longer have any limitations.

ScottCarda-MS: updated to remove the sentnece about the restriction.

Specifications/Language/3_Expressions/ConditionalLoops.md
ScottCarda-MS: removed

swernli (approved):

